PR #5, opened 2015-12-05
add version menu to download page

sblaisot:
Add a fixed-positionned menu on the left of download page (only if screen size > 1150px because we don't have space for that on smaller screens) to make 2.0-rc1 more visible to users clicking on the large download button of the homepage.


sblaisot:
getting closer to 2.0 makes this PR unneeded anymore. Closing it.

PR #6, opened 2015-12-17
push RC2

ywwg:


rryan:
LGTM -- you can push to staging by doing `fab clean build staging public`

PR #7, opened 2015-12-21
Frontpage rewrite

Be-ing:
Rewrite the front page to not come off as sensationalist and exaggerated


Be-ing:
Whoops, wrong branch

PR #10, opened 2015-12-22
Features page rewrite

Be-ing:
Rewrite features page to use more concise and down-to-earth language rather than sensational hype. Also shuffle some content around. Remove "Full specifications" section from already long page and move some information from it to relevant sections

PR #8, opened 2015-12-21
Frontpage rewrite

Be-ing:
Rewrite the front page to not come off as a sensationalist advertisement

PR #15, opened 2017-05-04
More HTTPS, fix mixed content

ishitatsuyuki:


Be-ing:
Thanks for bringing this up. I think this will likely cause merge conflicts with #8 and #10, so let's wait until those are merged before merging this.

sblaisot:
Why not using protocol agnostic links ?
https://www.paulirish.com/2010/the-protocol-relative-url/

use //domaine.com/filepath instead of https://domaine.com/path. This way, it will work with both http and httpd depending on the version you browse.

ishitatsuyuki:
Why using that? Loading HTTPS on HTTP is fine IIRC.

rryan:
> Why using that? Loading HTTPS on HTTP is fine IIRC.

We may need to disable HTTPS at some point in the future (e.g. during server maintenance or migration) and it would be nice to have that not break things.

daschuer:
Is this one still relevant? Can we close it?

ishitatsuyuki:
Mixed content is still a thing. Given that this is unblocked, should I rebase?

ishitatsuyuki:
Closing due to inactivity.

PR #18, opened 2018-01-04
replace IRC with Zulip

Be-ing:
also remove mention of the email list. At least until Zulip archives can be indexed by search engines, the email list is still linked on the [wiki](https://mixxx.org/wiki/doku.php).

PR #24, opened 2018-04-12
Replace IRC with Zulip.

rryan:
Continuation of #18.


pages/support.html
```diff
@@ -36,21 +36,14 @@ <h2>{% trans "Community Forums" %}</h2>
   </div>
 
   <div class="halfbox_right">
-    <h2>{% trans "Developer Mailing List" %}</h2>
```
daschuer, line 39: We we still need the mailing list for low traffic announcements

PR #25, opened 2018-04-15
update Downloads page for 2.1 release

Be-ing:
TODO: make release announcement on the forum

Do not merge until we have affirmative confirmation that the release builds are working for every OS and the builds have been moved to the download URLs linked here.


pages/download.html
```diff
@@ -206,7 +207,7 @@ <h2><img src="{% static '/static/images/download_mac.png' %}" class="feature_ico
     <p>
       <a href="{{ stable_osxintel }}"
          onClick="javascript:trackDownload('{{ stable_osxintel_analytics_conversion }}');">
-        {% blocktrans %}Download {{ stable_long_version }} for Mac OS X 10.5+ (Intel){% endblocktrans %}</a>
+        {% blocktrans %}Download {{ stable_long_version }} for macOS 10.5+ (Intel){% endblocktrans %}</a>
```
Be-ing: Is this still correct? What is the oldest version of macOS still supported?

esbrandt: > Is this still correct? What is the oldest version of macOS still supported?

10.8, see https://github.com/mixxxdj/mixxx/blob/master/build/osx/product_definition.plist

PR #31, opened 2018-06-14
Change download to  2.1.1

daschuer:


Pegasus-RPG:
LGTM. Have you tested that none of the links are broken?

daschuer:
Yes.

pages/download.html
```diff
@@ -16,20 +16,20 @@
 unstable_ubuntu64_analytics_conversion: /downloads/2.1.0-rc1-ubuntu64
 unstable_ubuntu_ppa_analytics_conversion: /downloads/2.1.0-rc1-ubuntu-ppa
 stable_short_version: Mixxx 2.1
```
Be-ing: This needs to be updated to 2.1.1 too.

daschuer: Why? It is still a 2.1 Version. We have also not had the minor version before.

Pegasus-RPG: We should have had the full version number before!

Be-ing: It is inconsistent to have 2.1 in some places but 2.1.1 in others.

daschuer:
Done. Looks now less pleasant to me, but OK.  
http://staging.mixxx.org/download/

PR #32, opened 2018-06-18
Artful to Bionic

kazakore:
The Ubuntu Artful deb packages do not install on Bionic so removed "(and Newer)" from the link text.

Changed stable_ubuntu_max_supported_version to Bionic as it is supported by the ppa with no issues.

It does claim all Ubuntu versions from Trusty right through to (now) Bionic are on the ppa as version 2.1.1 but this is not in fact correct. Maybe close enough to accurate to leave as is though?? Maybe worth considering if this should be clarified too.....


daschuer:
ok to test

daschuer:
LGTM, thank you.

PR #33, opened 2018-08-15
update footer copyright to 2018

sblaisot:
lp:1784789

### DIFF
--- a/templates/footer.html
+++ b/templates/footer.html
@@ -42,6 +42,6 @@
 </div>
 
 <div id="footer_legal">
-  Copyright &copy; 2015 Mixxx Development Team<br>
+  Copyright &copy; 2018 Mixxx Development Team<br>
   Logo by Paul Bloch
 </div>


Pegasus-RPG:
It would probably be better to change the footer to a php file and have it do `"Copyright 2001 - ".date('Y');` so we don't have to keep doing this. :)

sblaisot:
All website is plain HTML so no PHP is possible here I think. Or that would make us redesign all website to use PHP which is another amount of work.
Either we use SSI (.shtml) if supported/configured on web server, but I'm not sure that's simple or use javascript which doen't seems to me like a good solution.

I prefer to stick on static date until someone else port the whole website to php if wanted.
I already update dates every year, so one more is not a problem:
https://github.com/mixxxdj/mixxx/pull/838
https://github.com/mixxxdj/mixxx/pull/1101
https://github.com/mixxxdj/mixxx/pull/1452

Pegasus-RPG:
Okay, good point. LGTM.

PR #41, opened 2018-10-11
Remove links to google plus

radusuciu:
Now that google is shutting down google+, it makes sense to remove the ability to share on the network as it encourages users to share to FB/Twitter instead.


Pegasus-RPG:
ok to test

radusuciu:
Closing as was merged in with #42

Issue #11, opened 2016-07-28
website is not responsive


Be-ing:
The current website design does not work well with smartphone screens.

PR #42, opened 2018-10-12
Make website responsive

radusuciu:
Was browsing the website on my phone the other morning and noticed that it wasn't optimized for mobile as noted in #11.

Further testing is needed before merging (am happy to do this), and there are a few areas where things could be improved. At time of writing, the footer could squeeze down a bit better, and some images on the features pages have fixed widths.

The biggest change is that the menu squishes together in two stages, eventually collapsing down into a CSS-only hamburger menu. To avoid repeating the menu items and also to avoid changing the existing structure too much, menu items are now defined on the page context found in `/plugins/page_context.py`. As far as I can tell, this is fairly standard using Cactus. The site should be basically identical when viewed on a wide screen desktop computer, but here it is before and after on smaller screens (emulated Firefox 62)

### Before
screen shot 2018-10-12 at 15 01 44

### After
screen shot 2018-10-12 at 15 01 48

### Before
screen shot 2018-10-12 at 15 02 18

### After
screen shot 2018-10-12 at 15 02 23

### After, with open hamburger menu
screen shot 2018-10-12 at 15 02 29

### Before
screen shot 2018-10-12 at 15 02 39

### After
screen shot 2018-10-12 at 15 02 46

### After, with open hamburger menu
screen shot 2018-10-12 at 15 02 51

A very elegant, pure CSS, MIT licensed hamburger menu implementation was adopted from https://codepen.io/erikterwan/pen/grOZxx Vendor prefixes for some properties are required. We can also make it less fancy if desired. A JS implementation is also trivially possible, but I think sticking with CSS only is ideal.

### DIFF
--- a/pages/features.html
+++ b/pages/features.html
@@ -32,7 +32,7 @@
     </p>
   </div>
   <div class="halfbox_right">
-    <img width="362px" src="{% static '/static/images/feature_deck.png' %}">
+    <img class="responsive" width="362px" src="{% static '/static/images/feature_deck.png' %}">
     <p><b>{% trans "Beat Rolls and Censor" %}</b><br>
       {% trans "Play with rhythm by triggering short loops and reverse-playback effects. The track stays in time so you don't miss a beat." %}
       {# TODO: Write a better way of explaining slip mode. #}
@@ -57,14 +57,14 @@
       {% trans "Load up to 64 sampler decks full of sounds to layer over your mix." %}{# TODO: Say more. #}
     </p>
 
-    <center><img src="{% static '/static/images/feature_sampler2.png' %}"></center>
+    <img class="center responsive" src="{% static '/static/images/feature_sampler2.png' %}">
     
     <h2>{% trans "Effects Chains" %}</h2>
     <p>
       {% trans "Link up to three effects in a chain for a unique twist on your mix. Fine tune your sound by tweaking each parameter individually. Customize your workflow by assigning different parameters to adjust with the effect's metaknob. Focus different effects from a controller for flexible control of effect chains, whether your controller has a full effects section or a single knob." %}
     </p>
 
-    <center><img src="{% static '/static/images/feature_effect_unit_expanded.png' %}"></center>
+    <img class="center responsive" src="{% static '/static/images/feature_effect_unit_expanded.png' %}">
   <div style="clear: both;"></div>
   
     <h1>{% trans "Customizable DJ Hardware Support" %}</h1>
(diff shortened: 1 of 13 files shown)


Be-ing:
ok to test

Be-ing:
Awesome, thanks for working on this!

radusuciu:
Thanks! Will continue working, as it seems that the viewport is off on an actual phone and the hamburger menu ends up getting displayed at all times.

sblaisot:
Nice work, Thank you @radusuciu !

However, needing to discover and open the sandwich menu to find links to other pages is probably disturbing for users. Is it possible to keep direct links to pages without menu ?

radusuciu:
@sblaisot I agree, and I think I can squeeze it in for a tablet sized screen but if you look at the phone screenshot it's too small to fit anything without basically removing the much loved Mixxx logo.

We could try having the menu flow underneath the logo but I don't think it'd be very attractive and it would push down the main content as well. For those reasons, I think a hamburger menu solution is ideal for phone sized screens.

Having said that, I'm happy to implement whatever the consensus is!

Edit: I'll hop on Zulip and discuss more there

daschuer:
I had just a brief look, and it works already nice for me. Thank you.

radusuciu:
Ready to go from my perspective, please let me know if you find issues or would like to see changes. Changes since original submission:
- menu stays around for longer and only gives way to hamburger menu below 530 px. This is accomplished by making the font size smaller at smaller viewport widths. Compromise in response to request by @sblaisot. I think this makes sense
- images throughout the website are now responsive. This means that when the viewport width is smaller than the size of the image, they scale with their container.
- fixed issue with `overflow-x` property not being respected by some mobile browsers

Tested in latest versions of Chrome, Firefox, and Microsoft Edge (all three on mobile and desktop) as well as Internet Explorer 11 including Document Modes for IE10 and IE9 with no issues. Some issues observed with IE8 document mode with viewport widths below 1100px - the smaller menu entries don't appear and the hamburger menu doesn't work. I can address these if so desired (maybe in a separate pull request), but not worth it IMHO.

Please note that I based this off of #41, so google plus is removed. As stated in that pull request, I think this makes sense given that google+ is being retired soon, and also the site looks better without it IMO (though the design can be changed to accommodate this).

Be-ing:
This looks great, thanks so much! If you find more aspects to polish, please open a new pull request.

I'm at the GSOC Mentor Summit now and got compliments on the Mixxx website earlier from one of the MuseScore developers. :)

radusuciu:
Awesome, thank you, I'm happy to contribute! Can close #11!

PR #45, opened 2018-10-14
Fix issue where hamburger menu can be scrolled to 

radusuciu:
This affected some mobile browsers including Opera Mini and some versions of Safari. Fix is adding position relative to a containing element.

Please allow Jenkins to build test, and then for confirmation that the bug is resolved on affected browsers.

### DIFF
--- a/static/css/mixxx.css
+++ b/static/css/mixxx.css
@@ -35,12 +35,12 @@ body
 /* styles for general elements */
 a
 {
-  /* color: #d0d0d0; very light grey */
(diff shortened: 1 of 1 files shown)


Be-ing:
ok to test

rryan:
add to whitelist

Be-ing:
[Reported working on Zulip](https://mixxx.zulipchat.com/#narrow/stream/109122-general/subject/Responsive.20website/near/135787726). Thanks for the fix!

Issue #13, opened 2017-02-15
no link to twitter


ywwg:
People like the twitters, it's fantastic.  We should like to the twitters, where we have a very decent following

Be-ing:
We have a widget to share the page on Twitter, but a link to our Twitter account would be good too.

ywwg:
wordpress has a twitter widget that I use, maybe we can figure out an easy way to embed the latest activity on our twitter stream?  I try to like / RT mixxx-related things every week so it's not dead

sblaisot:
You can have a twitter widget on a static page too.
https://twitter.com/settings/widgets
just create the widget with the Mixxx account, copy/paste the javascript code in your static page and you're done.

sblaisot:
Here is the code you need to include in the page:
`
<a class="twitter-timeline" href="https://twitter.com/mixxxdj">Tweets by Mixxx</a> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>
`

radusuciu:
I think we can close this now!

PR #46, opened 2018-10-15
Replace social media widgets with social media links

radusuciu:
Removing social media widgets which are dynamically loaded has two main benefits:
- faster page loads
- better for privacy

The main difference is instead of clicking "Tweet" and being invited to tweet, you are linked to the Mixxx Twitter page (as requested in #13). In my opinion this is likely to increase engagement with Mixxx's social media. Would be nice if there was some data on how much use the widgets are getting.

The social media links in this PR are styled very similar to the widgets. Also removed jQuery as the only uses of Javascript on the website were simple enough to express without jQuery. It can always be added when needed.


ronso0:
I just now became really aware of the website repo..

I wanted to have a quick look at the icons (maybe apply Mixxx style?) but noticed the svgs are inline. does it have any benefit? do you mind storing the icons in separate files so they're easier to review and edit?

ronso0:
okay, got it up & running.
before I never was aware of those buttons as my browser blocks fb items..
it's nice to get rid of tracking objects and just link to the social sites instead!

as I thought in the first place, we could omit the text and just use icons.
Without text the buttons would also be small enough to fit underneath **Manual**, next to **Press**. This would make the site look a bit tidier than now.

radusuciu:
Svg as files would provide a minor benefit through caching I suppose, but it may be balanced out by increased number of requests (unless on HTTP2) - I think. I don't recall exactly but I think I just did it because it was easier.. can of course convert to files

I can pick this back up at some point but I'm currently writing my PhD Thesis so... lol. I'll be done in May but feel free to take over if you want!

Also, I should mention that I was thinking of splitting this up into two pull requests since I also cleaned up some JS.

Be-ing:
I just remembered about this PR. Thank you for reducing external dependencies and getting rid of the tracking. I noticed that the icons can overlap the "DJ Your Way" text:
image
compared to the current website:
image

How about getting rid of the "Twitter" and "Facebook" text and just using their icons?

Be-ing:
Also thanks for adding the Zulip button :+1:

ronso0:
yeah, that's pretty much what I proposed above and that would help in narrow view, too
> Without text the buttons would also be small enough to fit underneath Manual, next to Press. This would make the site look a bit tidier than now.

@radusuciu Are able to finish this or should someelse take over?

static/js/mixxx.js
```diff
@@ -0,0 +1,32 @@
+var OSName = getOSName();
+
+function getOSName() {
+  var OSName="your OS";
+  if (navigator.appVersion.indexOf("Win")!=-1) OSName="Windows";
```
Be-ing: Let's use the same style we use for JS in controller mappings:
```suggestion
  if (navigator.appVersion.indexOf("Win") !== -1) {
    OSName="Windows";
  }
```

Be-ing (changes requested):

ronso0:
yeah, that's pretty much what I proposed above and that would help in narrow view, too
> Without text the buttons would also be small enough to fit underneath Manual, next to Press. This would make the site look a bit tidier than now.

@radusuciu Are able to finish this or should someelse take over?

radusuciu:
@ronso0 @Be-ing I can take into account above suggestions and amend the PR!

radusuciu:
Reverted a commit that added these new side link icons to all pages, as some pages already had elements in the same position, and I think that that would warrant a separate PR.

radusuciu:
Other changes:
- moved inline svg to separate files
- re-organized js a bit. Namespaced what few functions there are and extracted some js from the download page to it's only file

ronso0:
Looks good on desktop and mobile view, works as expected.
Are there any corner cases to test?

radusuciu:
> Looks good on desktop and mobile view, works as expected.
> Are there any corner cases to test?

Not that I can think of other than trying at different screen sizes.

Be-ing:
Thanks for taking care of this.

PR #53, opened 2019-03-02
change copyright to 2019

ronso0:


daschuer:
https://epiphany.law/articles/copyright/copyright-dates-single-year-or-range
I think we need to use a range eg. 2000-2019

daschuer:
The written copyright and the sign are redundant. Use only one.

ronso0:
Re year/range
there are different opinions, and AFAICT the year of the first publication would be sufficient.
however, a range indicates that the project is maintained.
I'll change it to **&copy; 2001-2019 Mixxx Development Team**

daschuer (approved):
Thank you. LGTM

Pegasus-RPG:
In the future, we should just use `<?php echo date('Y');?>` so it is always up-to-date. That does require of course changing the file extension to .php and adjusting every place it's included.

ronso0:
> In the future, we should just use `<?php echo date('Y');?>` so it is always up-to-date. That does require of course changing the file extension to .php and adjusting every place it's included.

sure, what's necessary to implement this?

Be-ing:
I'd rather not introduce PHP to the static website. A little JS could take care of it though

sblaisot:
javascript in the browser is client-side. You have absolutely no idea of the date on the client browser. (well, at least not too far away or it will break TLS negociation).

It surely is good enough to print a copyright date on the site but it should be kept in mind as it could cause great problems in another context.

sblaisot:
Does the server support [Server-Side includes](https://en.wikipedia.org/wiki/Server_Side_Includes) ?

For example for [apache](https://httpd.apache.org/docs/current/howto/ssi.html) or [nginx](http://nginx.org/en/docs/http/ngx_http_ssi_module.html)

Simply add in your html file
```
<!--#config timefmt="%Y" -->
<!--#echo var="DATE_LOCAL" --> 
```

PR #65, opened 2019-11-05
remove Google Analytics

Be-ing:
This proprietary JavaScript collects data about website visitors
without their consent and sends it to an unaccountable third
party. Considering we haven't used the data from it in years, I
do not know why would need to now or in the future. If we really
want this data, we could setup Matomo on the mixxx.org server,
but I do not think this is worth the trouble.

### DIFF
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -12,12 +12,6 @@ unstable_win64: https://downloads.mixxx.org/builds/2.3/Windows/Mixxx-2.3-latest.
 unstable_macosintel: https://downloads.mixxx.org/builds/2.3/macOS/Mixxx-2.3-latest.dmg
 unstable_ubuntu32: https://downloads.mixxx.org/builds/2.3/release/mixxx-2.3.0-beta-2.3-release-bionic-i386-latest.deb
 unstable_ubuntu64: https://downloads.mixxx.org/builds/2.3/release/mixxx-2.3.0-beta-2.3-release-bionic-amd64-latest.deb
-unstable_win32_analytics_conversion: /downloads/2.3.0-beta-win32
-unstable_win64_analytics_conversion: /downloads/2.3.0-beta-win64
-unstable_macosintel_analytics_conversion: /downloads/2.3.0-beta-osxintel
-unstable_ubuntu32_analytics_conversion: /downloads/2.3.0-beta-ubuntu32
-unstable_ubuntu64_analytics_conversion: /downloads/2.3.0-beta-ubuntu64
-unstable_ubuntu_ppa_analytics_conversion: /downloads/2.3.0-beta-ubuntu-ppa
 stable_version: Mixxx 2.2.4
 stable_release_announcement: /news/2020-06-11-Mixxx-2-2-4-released/
 stable_win_min_version: 7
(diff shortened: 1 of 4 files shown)


Be-ing:
After this and #46, I think all nonfree JavaScript will be gone from mixxx.org

ywwg:
I know @rryan has used the analytics dashboard in the past to see how our site is performing and how people are coming in.  I also have access, though I haven't used it in a long time.  But let's doublecheck first that he doesn't still think it's worth having

daschuer:
Removing or not.  We should ask the user for permission.
This it required from a new EU privacy directive at least.

radusuciu:
I think that there are some alternatives as well. For example: https://matomo.org/matomo-on-premise/. We could still get some analytics this way without also beaming the data to google..

ywwg:
I checked, and I still have access to the mixxx analytics dashboard.  If we want to move to some new solution then we can export the data we already have so we don't lose that history.

Be-ing:
> I know @rryan has used the analytics dashboard in the past to see how our site is performing and how people are coming in.

Has this led to any decisions that would not have been made otherwise? Do we actually need this information?

ywwg:
if he doesn't chime in to say anything, then no and we can go ahead and remove it.

rryan:
Hm... 

It produces valuable data that we've used in the past to help inform decisions about what platforms to support (when can we axe Vista support?), resolutions to design for, languages to focus on, etc. I think we actually underutilize this data! It's really hard to build a product without data about who you're designing for.

I understand the concerns about non-free Javascript and 3rd party data -- but I think we should consider being pragmatic on this front. We already do a lot for the world by releasing Mixxx as Free software and IMHO it helps us achieve our mission to have this data. Also, anyone who is bothered by this is probably already blocking GA.

rryan:
Also, download metrics and referral sources (e.g. how much traffic was driven by that DJTT blog post?)

Holzhaus:
We could switch to Matomo, which would remove sending user data to a third party, however I think we'd need a GDPR disclaimer anyway.

Holzhaus:
@rryan Ping. If we stick with Google Analytics, can we at least anonymize user IP addresses? https://support.google.com/analytics/answer/2763052?hl=en

In Germany some lawyers are sending out cease and desist letters with a penalty clause to websites that do not have this enabled because it's a blatant GDPR violation.

ywwg:
That seems like an acceptable feature to turn on!  I don't think we care about individual ips

katsar0v:
Google Analytics is still on the website and we do not ask for any consent. I would also remove the analytics script or make a PR with a cookie consent popup (which is something we do not want to have now I guess)

Be-ing:
I redid this for the new Pelican website. We still haven't done anything useful with the data nor annoyed users with a GDPR popup. Now that we are using Netlify for hosting which only supports static sites without a database, using Matamoto is unfeasible. Let's get rid of this proprietary spyware already.

Be-ing:
Ping can we please get rid of this already.

Be-ing:
ping

Holzhaus:
I'm in favor of this.

*If* we keep Google Analytics we need to ask for consent before we may use it ( for GDPR compliance, see the [2019 rulings](https://www.kevel.co/blog/gdpr-google-analytics/) by the Conference of the Data Protection Authorities in Germany and the Court of Justice of the European Union (CJEU)). Last time we discussed this it was more than a year ago and considering that we didn't use or even look at the google analytics data at all, I'd say it's not worth it to add a popup to our website. @rryan What do you think?

rryan:
My stance is the same as stated here: https://github.com/mixxxdj/website/pull/65#issuecomment-552297359

I think decisions made by an organization are worse and more biased without data. 🤷‍♀️ 

Concretely, the most useful info is:
* OS / version distribution
* screen resolutions
* languages

It's true that simply recording the data is not enough to get the benefits. We would need a data-driven culture on our team beyond just having the data.

For a long while a huge component (possibly the majority) of our download stream was low-spec Windows machines in India. It would make me very happy if Mixxx worked really well for those users! We always run the risk of optimizing Mixxx towards only working well for people on high-end hardware and it's all too easy for us to have privilege-blinders on when we decide how to design our product.

Be-ing:
I don't think we need to surveil visitors to our website to be mindful of performance on slow hardware.

Holzhaus:
@rryan I agree that we should do more analysis on our target audience. However, we currently don't even though we have GA. Also, I don't think it's worth it having to add a Cookie popup to comply with GDPR. This is not a theoretical issue, there are lawyers that send out thousands of cease-and-desist letters and make people pay hundreds of euros because of this.

Also, the data collected here is not that useful, because it's heavily skewed towards Windows users (most Linux users will install Mixxx via their package manager and never visit our website). We could look into other ways to get some data on our users, e.g. user surveys, opt-in (!) telemetry, etc...

Be-ing:
Considering we were in blatant violation of GDPR, I went ahead and merged this. If someone thinks this is really important, then you can open a PR reintroducing analytics in a GDPR-compliant way which does not require proprietary JavaScript. I still think that would be annoying, unnecessary, and maintenance burden.

Be-ing:
image
:+1:

PR #72, opened 2020-02-21
pages: Show latest news on home page

Holzhaus:
Here's my attempt on adding news to the home page. It doesn't look great, so maybe someone with better web design skills wants to improve it a bit before merge.

2020-02-21-122112_1322x1069_scrot

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -66,6 +66,38 @@
     <div style="clear: both;"></div>
   </div>
 
+  {% if posts|length >= 2 %}
+  <h1>{% trans "Latest News" %}</h1>
+  <p class="header">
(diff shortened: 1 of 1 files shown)


Be-ing:
@ronso0 any thoughts about the design?

rryan (approved):
Cool! The only thing I find odd is the date being at the bottom instead of closer to the title. But IMHO we can tweak it in a future PR because it looks good as is.

ronso0:
Until I see the section in context I can't comment on the design, but I'm travelling until march.
Design tweaks can be applied after merge. 

From the preview I see I think the News should be in a more prominent place, maybe at the top?

rryan:
> Until I see the section in context I can't comment on the design, but I'm travelling until march.
> Design tweaks can be applied after merge.

SGTM, I think this is good to go.

PR #67, opened 2020-02-14
Add News pages

Holzhaus:
The [blogspot site](https://mixxxblog.blogspot.com/) linked on the website is unmaintained.

IMHO it makes more sense to have this integrated into the main website.

I added support for a news page and migrated the old blog posts using the Atom feed. I also added the newer release announcements from the forum posts.

What's still missing is an RSS/Atom feed, but we can add this later on. If this is merged, someone could add a post to the old blog that links to the new news page.

PR #73, opened 2020-02-23
use international y-m-d format for news dates

daschuer:
This is a solution for the ambiguous dates issue discussed here: 
https://github.com/mixxxdj/website/pull/67

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -78,7 +78,7 @@
     <p class="feature_indent">{{ post.post|truncatewords_html:30 }}</p>
     <p>
       <a href="{% url post.path %}">{% trans 'Read More' %} &raquo;</a>
-      (<time pubdate="pubdate">{{ post.date|date:"SHORT_DATE_FORMAT" }}</time>)
+      (<time pubdate="pubdate">{{ post.date|date:"Y-m-d" }}</time>)
     </p>
     {% endwith %}
   </div>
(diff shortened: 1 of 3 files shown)


rryan (approved):
Please also update 
https://github.com/mixxxdj/website/blob/d2158047f070afdc214e0e7c1ff629a000f7079f/pages/index.html#L75-L95
and 
https://github.com/mixxxdj/website/blob/d2158047f070afdc214e0e7c1ff629a000f7079f/templates/post.html#L13

Holzhaus:
I had hoped that someone knows how to change the value of `SHORT_DATE_FORMAT` in the settings, so that we don't have to edit all templates manually, but since that's not the case we need to go this route I guess.

rryan:
> I had hoped that someone knows how to change the value of `SHORT_DATE_FORMAT` in the settings, so that we don't have to edit all templates manually, but since that's not the case we need to go this route I guess.

I think this controlled by Django localization. If we were to build the website with locale en-GB I think `SHORT_DATE_FORMAT` would automatically become `d/m/Y`. I don't know how to set it manually.

daschuer:
Done. Merge?

Holzhaus:
What about the index page?

Holzhaus:
> What about the index page?

https://github.com/mixxxdj/website/blob/website/pages/index.html#L81

daschuer:
Done. I had to merge from master first to see the changes. 
Merge now?

Holzhaus (approved):
Thanks, LGTM.

PR #75, opened 2020-03-25
Warn users about fake paid versions of mixxx

ywwg:


pages/index.html
```diff
@@ -15,7 +15,7 @@ <h1>{% trans "DJ Your Way" %}</h1>
 
 {% block body %}
 <div class="gapfiller" style="margin-bottom: 0px;">
-  <p>Free and open source DJ software for Windows, macOS, and Linux</p>
+  <p><strong>Totally Free</strong> and open source DJ software for Windows, macOS, and Linux</p>
```
Be-ing: I don't think we need this extra emphasis.

pages/download.html
```diff
@@ -212,8 +212,10 @@ <h1>{% blocktrans %}Download Now ({{ stable_version }} Stable){% endblocktrans %
   <div class="fullwidth">
     <p>{% blocktrans %}Check out the <a href="{{ stable_release_announcement }}">release announcement</a> for a list of new features.{% endblocktrans %}</p>
     <p>{% blocktrans with previous_link="#previous" %}Although {{ stable_version }} contains many important updates and bugfixes, it's possible some users may experience issues. If you do, you can still get our previous stable version, <a href="{{ previous_link }}">{{ previous_version }}</a>.{% endblocktrans %}</p>
-    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}<br/>
-    {% trans "We happily provide Mixxx for free and donations are appreciated!" %}</p>
+    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}</p>
+    <h3>{% trans "If you ever see Mixxx being offered for sale, it is an unauthorized distribution." %}<br/>
```
Be-ing: This isn't necessarily true. By publishing software under the GPL we explicitly authorize anyone to sell it so long as they provide source code. Most of the time they probably don't provide source code or even a link to the source code, but there could be someone out there selling Mixxx that actually is complying with the GPL. If there isn't now, there could be in the future. How about something like:

Mixxx is available for free for Windows, macOS, and Linux. Donations are appreciated but not required.

If you see Mixxx offered for sale on other websites, there is no need to pay for it because you can get it here for free. There is no "pro" version of Mixxx with extra features; the free version is the full version.

ywwg: we can still say that those versions are not "official" though.  Also they shouldn't be using our name

ywwg: updated text

pages/download.html
```diff
@@ -212,8 +212,11 @@ <h1>{% blocktrans %}Download Now ({{ stable_version }} Stable){% endblocktrans %
   <div class="fullwidth">
     <p>{% blocktrans %}Check out the <a href="{{ stable_release_announcement }}">release announcement</a> for a list of new features.{% endblocktrans %}</p>
     <p>{% blocktrans with previous_link="#previous" %}Although {{ stable_version }} contains many important updates and bugfixes, it's possible some users may experience issues. If you do, you can still get our previous stable version, <a href="{{ previous_link }}">{{ previous_version }}</a>.{% endblocktrans %}</p>
-    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}<br/>
-    {% trans "We happily provide Mixxx for free and donations are appreciated!" %}</p>
+    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}</p>
+    <h3>{% trans "You don't need to pay for Mixxx because you can get it here for free." %}<br/>
+        {% trans "<strong>There is no official paid or &quot;pro&quot; version of Mixxx.</strong>" %}<br/>
```
Be-ing: I don't think we need `<strong>` here. For most people looking at this page, this isn't important.

ywwg: removed

pages/index.html
ywwg: done

Be-ing:
Hmm the paragraph still looks emphasized:
image

pages/download.html
```diff
@@ -212,8 +212,11 @@ <h1>{% blocktrans %}Download Now ({{ stable_version }} Stable){% endblocktrans %
   <div class="fullwidth">
     <p>{% blocktrans %}Check out the <a href="{{ stable_release_announcement }}">release announcement</a> for a list of new features.{% endblocktrans %}</p>
     <p>{% blocktrans with previous_link="#previous" %}Although {{ stable_version }} contains many important updates and bugfixes, it's possible some users may experience issues. If you do, you can still get our previous stable version, <a href="{{ previous_link }}">{{ previous_version }}</a>.{% endblocktrans %}</p>
-    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}<br/>
-    {% trans "We happily provide Mixxx for free and donations are appreciated!" %}</p>
+    <p>{% trans "Mixxx is available for Windows, macOS, and Linux." %}</p>
+    <h3>{% trans "You don't need to pay for Mixxx because you can get it here for free." %}<br/>
+        {% trans "There is no official paid or &quot;pro&quot; version of Mixxx." %}<br/>
+        {% trans "The version available here is the full version." %}</h3>
```
Be-ing: remove the `<h3>`

Holzhaus: Yup. Even if we decide to keep the emphasis we should remove the `h3` tag as it denotes a headline, not emphasized text. HTML should be used semantically, so that users that depend on a screen reader can also use it properly.

ywwg: ok but the text was too small just as paragraph text.  what tag do you want me to use?

Be-ing: I think it should be paragraph text like the rest of the text surrounding it.

ywwg: I don't think it's prominent enough, but ok

Holzhaus:
You can use strong IMHO, but the current state is okay too.

Be-ing:
thanks

PR #87, opened 2020-05-10
Add CSS-only gallery to frontpage

randombyte-developer:
https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Website.20splash.20image.20gallery

This is a quick attempt without any JS.
Clicking on a thumbnail image navigates the browser to the regarding `#` anchor marker. This causes the large image container to scroll to the correct position to show the image.
I couldn't figure out how to center the thumbnails while keeping the ability to correctly scroll horizontally through the thumbnails on smaller screens. Input is appreciated.

`justify-content: center`, `margin: auto` and `width: fit-content` in various combinations either prevented the scrolling or didn't allow to scroll completly to the left.

This uses the screenshots I found were used on the feature site.

Large screen:

Smaller screen:
127 0 0 1_8000_(Galaxy S5)

PR #89, opened 2020-05-14
Added gallery to frontpage

randombyte-developer:
This is a rework of #87.
The gallery cycles through the four images I found in this repo, with each image having 5 seconds. Once the user clicks on the gallery the autoplay stops. The arrows on the gallery don't have to be clicked exactly: Clicking on the left halve of the whole image navigates you one back, right halve is next image.
I had some fun with svg and created the arrows myself.
This is the most minimal state of how it could be implemented, you could add sliding transitions, titles, page indicators, fullscreen images, etc.

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -4,12 +4,41 @@
 
 {% block title %}{% trans "Free DJ Mixing Software App" %}{% endblock %}
 
+{% block extra_script %}
+<script type="text/javascript" src="{% static '/static/js/gallery.js' %}"></script>
+<script type="text/javascript">
+  const images = [
+    "{% static '/static/images/feature_skins_deere.png' %}",
+    "{% static '/static/images/feature_skins_latenight.png' %}",
+    "{% static '/static/images/feature_skins_shade.png' %}",
+    "{% static '/static/images/feature_skins_tango.png' %}"
+  ];
+
+  // Unhide the navigation buttons. Those are hidden by default if JS is disabled and no navigation is possible.
+  // The buttons itself don't have a click listener, they just guide the user to click on the image.
+  const previousImageButton = document.getElementById("previous_image_button");
+  const nextImageButton = document.getElementById("next_image_button");
+  previousImageButton.style.display = "initial";
+  nextImageButton.style.display = "initial";
+
+  const splashGallery = document.getElementById("splash_gallery");
+  const splashImg = document.getElementById("splash_img");
+  setupGallery(splashGallery, splashImg, images);
+</script>
+{% endblock %}
+
 {% block after_navbar %}
 {% include "social_sidebar.html" %}
 
 <div id="splash">
-  <h1>{% trans "DJ Your Way" %}</h1>
-  <img class="center responsive" src="{% static '/static/images/splash-2.1.png' %}">
+  <h1 class="center">{% trans "DJ Your Way" %}</h1>
+
+  <div id="splash_gallery">
+    <!-- Setup the gallery in a way that still makes sense even if JS is disabled: Set default image and hide navigation buttons (displayed by JS again) -->
+    <img id="splash_img" src="{% static '/static/images/feature_skins_deere.png' %}">
+    <img id="previous_image_button" class="gallery_navigation_button" style="display: none;" src="{% static '/static/images/arrow_left.svg' %}" />
+    <img id="next_image_button" class="gallery_navigation_button" style="display: none;" src="{% static '/static/images/arrow_right.svg' %}" />
+  </div>
 </div>
 {% endblock %}
 
--- a/pages/index.html
+++ b/pages/index.html
@@ -7,13 +7,6 @@
 {% block extra_script %}
 <script type="text/javascript" src="{% static '/static/js/gallery.js' %}"></script>
 <script type="text/javascript">
-  const images = [
-    "{% static '/static/images/feature_skins_deere.png' %}",
-    "{% static '/static/images/feature_skins_latenight.png' %}",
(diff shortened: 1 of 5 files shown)


Be-ing:
add to whitelist

pages/index.html
```diff
@@ -4,12 +4,41 @@
...
 
+{% block extra_script %}
+<script type="text/javascript" src="{% static '/static/js/gallery.js' %}"></script>
+<script type="text/javascript">
+  const images = [
+    "{% static '/static/images/feature_skins_deere.png' %}",
+    "{% static '/static/images/feature_skins_latenight.png' %}",
+    "{% static '/static/images/feature_skins_shade.png' %}",
+    "{% static '/static/images/feature_skins_tango.png' %}"
+  ];
```
radusuciu: Can this be avoided? Or is hardcoding this not reliable? I believe urls like these are already hardcoded in the css, so I would consider moving this initialization code into `gallery.js`

static/js/gallery.js
```diff
@@ -0,0 +1,26 @@
+function setupGallery(galleryContainer, imgElement, images) {
```
radusuciu: I would wrap the JS with an IIFE like this: https://github.com/mixxxdj/website/blob/website/static/js/download.js, exposing a namespace like this: https://github.com/mixxxdj/website/blob/website/static/js/mixxx.js

Though I suppose since you're using block scope (which the other code code not do) this is less of an issue?

radusuciu (commented):
Implementation is sound and works. Made some suggestions on the code.

Issues:
- responsiveness is broken
- size is off. compare front page now and with this PR:
image
image

Other suggestions:
- Add a css transition effect

pages/index.html
randombyte-developer: I wanted to keep `gallery.js` as general as possible to be reused elsewhere. That's why those URLs are not in `gallery.js`. I can put them in a separate JS file. But those files might only be temporary ones for the first test.
The original image has some kind of "border darkness" effect, which we might want to have again on the frontpage. 

At the location where those images are used in the first place (features.html) they are hardcoded in the HTML. Should the URLS be moved to a separate JS?

static/js/gallery.js
randombyte-developer: I deliberately decided against an IIFE since we have `let` and `const`now.

randombyte-developer:
I noticed that the images are not cached and always re-GET-ed when navigating through the gallery.

randombyte-developer:
Before the merge we need higher res images. I also don't know how up-to-date those are. Personally I only use Deere. Anyway, screenshots would have to be retaken. Based on Mixxx 2.2.3?
Any preferences which tracks should be visible if I take the screenshots?

Be-ing:
We should also have images in different scale factors and use srcset.

Be-ing:
This is better than #87. Can you get the transitions to use a scrolling animation?

randombyte-developer:
I added a fade animation. It already was a big hassle to get the dimensions right because the img elements are now absolute and need a specific width & height. I had to add a bunch of CSS classes to make it not break in a non-JS environment. The layout corrects itself on smaller screens only when reloading the page. This is not a problem on a real device (tested on my smartphone) but only has this reload-quirk on the browser web tools.
I tried to do a sliding animation but I had layout problems. If we want to have that animation I would definitely use a dedicated library.

radusuciu:
Shouldn't the image dimensions not change? Also, you can apply css-transitions if you use the background-image property of a containing element.

randombyte-developer:
The dimensions change on smaller screens. Automatic resizing doesn't work because the gallery has a fixed size because all gallery children are `position: absolute`. That's why I recalculate the gallery height in JS. I haven't found a way to do this in CSS only.
If the height is not adjusted, the navigation elements stay at the same height, because they use `top: 50%`. The parent gallery doesn't change its height, which means that the 50% are always the same, no matter of the screen size.
I hope this was understandable.

I haven't looked in background-image and animating that, but it also seems that resizing it with the navigation elements needs more logic than just plain CSS.
Any ideas?

radusuciu:
I believe it's doable, but like 60% certain lol. I'll take a look tomorrow

Edit: did not, but likely will need a rework due to recent redesign

Be-ing:
@randombyte-developer could you update this for the redesigned website?

Holzhaus:
Do we still want a gallery on the front page? I'm not sure that it's a good idea, at least a full-blown slider. Maybe a small row of thumbnails that you click on.

randombyte-developer:
I am also not sure about that with the new design. Maybe just a more prominent "Screenshots" button to https://www.mixxx.org/screenshots/ ?

Be-ing:
Sure that would work too.

Swiftb0y:
But before we link that page more prominently, we should definitely update it. Eg more pictures of the skins in different configurations at an adequate resolution.

Be-ing:
Shall we wait for https://github.com/mixxxdj/mixxx/pull/2913 to be merged before that?

Holzhaus:
> But before we link that page more prominently, we should definitely update it. Eg more pictures of the skins in different configurations at an adequate resolution.

I'm waiting until @ronso0 until he's ready with the LateNight fixes, then I can take care of that.

ronso0:
don't wait for mixxxdj/mixxx#2913, the changes there are minor and no one will notice a difference.

randombyte-developer:
So this PR can be closed since we will link to the screenshots page, right? In a new PR the screenshots have to be updated and the screenshots button would have to be placed somewhere better discoverable?

PR #96, opened 2020-05-22
Build website on Travis CI

Holzhaus:
This will make us less dependent on the build server (and should also free up the build queue a bit).

- [X] Build website on Travis CI
- [ ] Enable Travis CI for this repository
- [ ] Add deploy configuration, e.g. using rsync (@rryan I need SSH access to the server though. Maybe you can generate an SSH keypair that only allows access to the webserver docroot?)
  We can then [encrypt the SSH key](https://docs.travis-ci.com/user/encrypting-files/) and only decrypt it when a master build passed and we want to deploy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+python:
+- 3.7
+os: linux
+dist: xenial
+cache:
+  pip: true
+
+install:
+- python -m pip install -r requirements.txt
+
+script:
+- cactus build --verbose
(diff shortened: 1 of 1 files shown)


Holzhaus:
Can someone with sufficient right enable Travis for this repo and merge this? Then we can at least catch errors in templates and the python code early on.

Holzhaus:
Closing this in favor of #98.

PR #99, opened 2020-06-07
replace splash image with a controller that Mixxx supports

Be-ing:
photo from
https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/mixxx.20in.20the.20wild

Screenshot from 2020-06-07 11-06-26


ronso0:
Even though I thought the same earlier (we should show a controller we suport) it is IMO more important what mood we transport. The image from Zulip you propose has too much tech detail, and doesn't feel very lively. 
If we don't shoot a foto ourselves (which no one took care of the last years) let's not rush just for the sake of having a Mixxx sticker in the foto

If you need to replace the current foto because of the CDJs, there are those two on pexels.com we can use instead (cropped of course):
https://www.pexels.com/de-de/foto/beleuchtung-bunt-dj-konsole-farbenfroh-243870/
image

https://www.pexels.com/de-de/foto/auffuhrung-auftritt-band-beleuchtung-219101/
image

katsar0v:
I would add overlay, so the text and the logo are more visible. Maybe black with 15% opacity. What do you guys think?

daschuer:
Can we Photoshop the manufacturer logos away? This may leads to the impression that we are bounded to them in a way.

Be-ing:
I think it's okay as long as those logos aren't very prominent. The Hercules and Pioneer logos are prominent in both of the photos @ronso0 posted.

Be-ing:
Maybe we could take a photo with a Mixxx sticker covering a manufacturer's logo

ywwg:
I know this is super old but I actually like it a lot.  I know we updated the website to blur the header but how about we drop this in?

PR #101, opened 2020-06-07
Fix Audacity link on the maintenance page

### LINKED ISSUE #84: Audacity link is outdated on maintenance page
On the maintenance page is a link to Audacity. Apparantly they moved from http://www.audacity.org to https://www.audacityteam.org/. This should be updated.

Holzhaus:
Resolves #84.


daschuer (approved):
LGTM, thank you.

PR #105, opened 2020-06-10
make 'Free & Open Source' highlighted

katsar0v:
Small change, but this is also my first PR for this repo. I made this section highlighted:

image

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -79,7 +79,7 @@
   </div>
 </div>
 
-<div id="opensource" class="docs-section container">
+<div id="opensource" class="docs-section highlighted container">
   <h2><a class="headline-link" href="#opensource">{% trans "Free &amp; Open Source" %}</a></h2>
   <p>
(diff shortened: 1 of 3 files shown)


Be-ing:
add to whitelist

Holzhaus:
Thanks. This improves visibility of the donate button. However, I'm not sure about the highlighted section borders and background. How about using a background cover image like we do on the features page instead?

Something like this?
2020-06-10-124922_2637x1585_scrot

Holzhaus:
https://github.com/mixxxdj/website/pull/107

katsar0v:
@Holzhaus am not sure about the photo, but definitely a better highlight! If you could highlight the "donate" button in your PR (with the orange border color), I would vote for it

Does it make sense to move "donate" after download? It would be the second highlight

Holzhaus:
>  If you could highlight the "donate" button in your PR (with the orange border color), I would vote for it

Ok, I applied your change on my branch.

> Does it make sense to move "donate" after download? It would be the second highlight

Good idea, done in #107.

katsar0v:
Thank you @Holzhaus

PR #114, opened 2020-06-29
fix bad release note link

ywwg:


Be-ing:
Where did the June 11 date come from??

PR #115, opened 2020-06-30
pages/news/2020-06-29-black-lives-matter: Remove duplicate BLM headline

Holzhaus:
The hashtag is already in the post title and it looks a bit weird right
now.


ywwg:
LGTM

PR #116, opened 2020-07-02
change forum links to https://mixxx.discourse.group/

Be-ing:


Holzhaus:
We also need a rewrite rule that redirects all request for `/forum/...` to `https://mixxx.discourse.group/...`. Links should still work that way (needed for controller mapping links).

Be-ing:
Can we merge this first then take care of that when we move mixxx.org to Netlify?

Be-ing:
No point bothering to write an Apache rewrite rule when we'll have to redo it for Netlify anyway

Holzhaus:
> Can we merge this first then take care of that when we move mixxx.org to Netlify?

No idea when this will happen. Anyway, we can always do a follow-up PR and change the apache htaccess file later. Merge?

Holzhaus (approved):
LGTM.

PR #118, opened 2020-07-08
reorganize navigation menu

Be-ing:
* Rename "Support & Community" to just "Support"
* Add "Get Involved" top level item and add link to Zulip under it called "Chat"
* Rename "Discover" to "About"


Be-ing:
ping

Holzhaus:
> * Rename "Discover" to "About"

IIRC this was the original name and I was asked to rename it to "Discover". :shrug:

Be-ing:
"About" is a conventional category in a website navigation menu. It's where I'd expect to find what's in that menu.

Holzhaus:
This PR breaks the responsive navigation menu on my phone. It's too long and I can't scroll down.

Be-ing:
I'll look into that. Do you have any comments about the new organization?

plugins/page_context.py
```diff
@@ -19,17 +19,22 @@ def preBuildPage(page, context, data):
...
-        MenuItem("/support", "Support & Community", "Navigation bar link to Mixxx support page.", children=(
+        MenuItem("/support", "Support", "Navigation bar link to Mixxx support page.", children=(
             MenuItem("/manual/latest", "Manual", "Navigation bar link to Mixxx Manual."),
-            MenuItem("https://mixxx.discourse.group/", "Forums", "Navigation bar link to Mixxx Forums."),
             MenuItem("https://github.com/mixxxdj/mixxx/wiki", "Wiki", "Navigation bar link to Mixxx Wiki."),
-            MenuItem("/get-involved", "Get Involved", "Navigation bar link to Mixxx Get Involved page."),
+            MenuItem("https://github.com/mixxxdj/mixxx/wiki/Hardware%20compatibility", "Compatible Hardware", "Navigation bar link to Mixxx supported hardware wiki page."),
+            MenuItem("https://mixxx.discourse.group/", "Forums", "Navigation bar link to Mixxx Forums."),
+        )),
+        MenuItem("/get-involved", "Community", "Navigation bar link to Mixxx Community page.", children=(
```
Holzhaus: Should we rename this to "Get Involved" or "Development"? Otherwise I'd expect to find the forum here, but it also belongs into support and duplicating it would be weird.

Be-ing: We make a point on this page to invite people who aren't programmers, so I'm hesitant to rename it "Development".

Holzhaus: I like "Get Involved" better anyway. Or maybe "Contribute" to make it more consistent (all other top level nav items are one word now).

Be-ing: I like "Contribute"

daschuer (changes requested):
> Rename "Discover" to "About"

"About" does not work for me. It look boring and contains much more than we have in About menus in other contexts or even in Mixxx.  
What was the issue with "Discover"?

Be-ing:
Contact and Press don't make sense under a heading of "Discover".

daschuer:
I did not expect to find the "manual" under support, can we move this to "Discover"?

Be-ing:
I think Support is still the best header for the manual. I intentionally listed it above the forum link so users are more inclined to look for the answer to their question before posting.

daschuer:
> "About" is a conventional category in a website navigation menu. It's where I'd expect to find what's in that menu.

That is only true for websites like Github, where a web service is the main purpose. 
Our whole website is the About menu compared to Github. This matches the About menu in Mixxx.
Having another about menu on the Webside is redundant.

Be-ing:
> That is only true for websites like Github, where a web service is the main purpose.

No, here are some examples:
https://www.rane.com/about
https://www.kdevelop.org/kdevelop-team
https://nodejs.org/en/about/
https://www.ableton.com/en/about/

daschuer:
I have just discovered that About is clickable. This happens because It is not instantly highlighted as link. 
I am afraid many users will hit this trap. 

You are right that the context that is reachable with About matches About everywhere else. 

How about Use discover as a non-linked menu and move the About into the fly out? 

The same issue is valid for all other menus.

Be-ing:
> I have just discovered that About is clickable. This happens because It is not instantly highlighted as link.

Yes this is not easily discoverable. I think this is a styling issue that could be improved in another PR.

Be-ing:
> How about Use discover as a non-linked menu and move the About into the fly out?

I still think "Discover" is a strange header for Press and Contact.

daschuer:
The main issue is that we have "About" which stands for a usually boring text about a company also as headline for more interesting things. I don't mind to drop "Discover" but "About" should be a menu topic not the headline. 
Take "Features" if you wish. But I actually prefer not to hide own content under the menu group and find a catchy headline instead.

Holzhaus:
> > I have just discovered that About is clickable. This happens because It is not instantly highlighted as link.
> 
> Yes this is not easily discoverable. I think this is a styling issue that could be improved in another PR.

We can underline the text on hover.

Holzhaus:
> The main issue is that we have "About" which stands for a usually boring text about a company also as headline for more interesting things. I don't mind to drop "Discover" but "About" should be a menu topic not the headline.
> Take "Features" if you wish. But I actually prefer not to hide own content under the menu group and find a catchy headline instead.

TBH I think "About" fits quite well. "Features" is misleading when it contains press and contacts.

pages/news/2020-05-22-you-dont-need-to-pay-for-mixxx.html
```diff
@@ -59,7 +59,7 @@
 In fact, all our currently active developers started out as regular Mixxx users — just like you.
 
 Do you want to contribute to Mixxx, too?
-[You can!]({% url "/get-involved.html" %})
+[You can!]({% url "/contact.html" %})
```
Holzhaus:
```suggestion
[You can!]({% url "/contribute.html" %})
```

Holzhaus (changes requested):

katsar0v:
Here is my opinion as a web developer:

* When people see "About" they click to learn more. If there is not enough information on the current page, they usually click this link to learn more about a product or a service. Organisations put this link more often, just to present themself better. There is also another aspect of this page - seo. SEO plays a major (indeed) point when it comes to discovery via google - people look for your website, you have your "about us" page well optimized and they discover you quickly. Usually people optimize the about page, with microformats, proper text, well documented images and so on. From seo perspective this makes sense I think. From UX persspective it also makes sense to have a detailed page where you learn more about the product/software/service. The homepage should be a catching overview page(not overflowing information, because users might exit)

* @Holzhaus is right when he wrote that "Features" is misleading when it contains press and contacts. "Features" (from UX and SEO perspective) means that someone wants to learn more about what this product/service can do. If we do any kind of "advertisement" with press and contacts, it is basically not needed and should be moved to the very bottom of the page. I am proud with the [Features](https://mixxx.org/features/) page, but imo it can be extend. People who click or come on this page want to learn more about the features. We currently have 5 sections, but Mixxx can do more than that.

Be-ing:
Another useful PR killed by bikeshedding :/ The website has been redone since this was opened. Resuming this would require starting from scratch. I am not volunteering to do so.

ronso0:
We may pick single patches to improve the situation step by step.

Be-ing:
No, this PR was from before the switch to Pelican. It would need to be redone from scratch.

ronso0:
I meant that small patches can easily be reproduced with the new system, and reduce debatable changes > accelerate progress ; )

Be-ing:
Unfortunately it is not so easy to predict what changes will trigger bikeshedding. I would never have imagined changing "Discover" to "About" is remotely controversial.

ronso0:
maybe because Discover currently has mixed content
In **About** I'd expect
* Screenshots
* Press
* Contact

And have **Features** as single link in the navbar

but I don't necessarily want to restart the entire discussion

PR #119, opened 2020-07-08
revise front page header text

Be-ing:


Holzhaus:
I'm unsure if cross-platform is most attractive point. Sure, for Linux users it probably is, but I guess most of them already know that it works on Linux when the decide to look at our website. And "Cross-platform DJ Software" does not sound super appealing, at least for Windows/MacOS users. 

Maybe we should emphasize customizability or that it's completely free instead. 
E. g.: "The customizable, free DJ software" or "DJ how YOU like"  or something like that.

Be-ing:
I tried "Free and open source DJ software", but that's redundant with the "Free and open source" section right below that.

Be-ing:
I also tried moving "DJ Your Way" to below the navbar but I didn't like how that looked.

PR #122, opened 2020-07-09
Make Mixxx News section more blog-like (better overview page and add comments)

Holzhaus:


Holzhaus:
OK, I think this is ready now. I already added the proper CSS rules for the comments frame on the discourse side.
I made it possible to disable comments for individual posts. I also disabled comments for old posts to avoid flooding the forum with comment topic with old posts.

For testing, I made a news category that is only visible to staffers. We can either make that one visible or switch over to the "Announcements" category instead. All this can be done via the discourse interface, so my work on this PR is done.

I also made the news overview page more blog-like, which now lists the 4 latest posts with short excerpts (the older post are still a flat list). This makes it less confusing what page you're on.

PR #129, opened 2020-07-18
show dropdown icons in menu

katsar0v:
I think we should show some arrows:

image

Icon from: https://www.iconfinder.com/icons/186410/arrow_left_previous_icon


ronso0:
SVG hint:
in Inkscape, save as 'Plain SVG' reduce the file size by omitting all lines specific to Inkscape.

ronso0:
Good idea!
Can you color the arrows like the link fonts? And maybe lower the arrows a bit so they're centered vertically?

katsar0v:
New screenshot:

image

Holzhaus (approved):
LGTM, thank you!

PR #130, opened 2020-07-18
proposal: add info if using conda

katsar0v:
This is just a suggestion to improve the documentation, because of `cactus`.


Holzhaus:
This is not necessary anymore since we're now using Pelican which supports Python 3. Thanks.

PR #131, opened 2020-07-20
Fixed dead wiki links

wraybowling:
Links to the wiki were dead because wiki titles are case-sensitive


Be-ing:
Thanks for taking care of this!

PR #148, opened 2020-09-08
plugins/news: Make post time optional and fall back to "now" on error

Holzhaus:


Holzhaus:
This is a minor fix to make branches without a valid data build. A warning will still be displayed as before. Merge?

Holzhaus:
@Be-ing Merge?

ronso0:
LGTM, works as intended. (but I have no py knowledge unfortunately..)

Holzhaus:
> (but I have no py knowledge unfortunately..)

basically it just tries to parse the date in the `YYYY-MM-DD HH:mm:ss` format, and if that throws an error it tries `YYYY-MM-DD`, and if that also throws an error it just uses the current date and logs a warning.

ronso0:
... I thought so but I have no idea what can go wrong.

If you consider the change trivial and rather unlikely to cause a regression I can of course merge it now.

Holzhaus:
I think it's unlikely that this causes a regression, yes.

PR #151, opened 2020-09-23
Outreachy post

ywwg:


plugins/news.py
```diff
@@ -36,7 +36,7 @@
         'github': 'Pegasus-RPG',
     },
     'ywwg': {
-        'name': 'Owen Wilson',
+        'name': 'Owen Williams',
```
ywwg, line 39: I am not a celebrity :P

daschuer:
Thank you. I think that can make use of a paragraph how much of the donated money will reach the participant. 
Hopfully we can clarify the situations for Germany and other countries as well. I am waiting for responses to my questions.

ywwg:
> Thank you. I think that can make use of a paragraph how much of the donated money will reach the participant.

I replied with a general statement because the exact amounts will vary depending on how they donate.  Is that sufficient?

daschuer:
The 5 % For OpenCollective + 5 % for the fiscal host are constant. Than we should add a warning that a trans-fair via an non US bank is exorbitant expensive and should not be used. The PayPal fee is known. 
And we need to add a warning that PayPal charges silently + %5 due to the currency conversion.  
We should also state that the donation is not Tax deductible.

ywwg:
clarified values.  I don't want to list every single number because it will be different for everyone.  And rather than say "don't use paypal" I'd prefer to say "try to use a bank transfer"

daschuer:
From Germany PayPal is the only option.
In my case my Bank charges 8 € + 0.5% 
The credit card was resected.
In addition the website open collective website is broken for bank transfer. See above.

daschuer:
What exactly was the benefit of opencollective?
We can't make use of Tax deduction from anywhere?
Did you consider to move to betterplace.me?
This is cheaper and transferring the money to outreachy only costs 8€ + 0.5% one time.

ywwg:
Outreachy recommended opencollective specifically.  I don't know German and I'm in the US, so I can't use a non-US fundraising system.  As I said before, if you want to do the work to run a parallel European fundraiser that is fine with me. 

I don't believe we should be spending this time optimizing for the lowest collection fee when the internship application period starts on thursday.  Opencollective is giving me the support I need and has worked with Outreachy.  They are a good organization and I am comfortable with supporting them.

daschuer:
> clarified values. I don't want to list every single number because it will be different for everyone. And rather than say "don't use paypal" I'd prefer to say "try to use a bank transfer"

Did you forget to push? 

I think we need a sentence like: use bank transfer from US if possible. Use PayPal from other countries, but be aware of the extra 5% charge due to currency conversion.

ywwg:
oh yeah pushed to the wrong repo

pages/news/2020-09-23-outreachy.html
```diff
@@ -40,9 +40,14 @@
 You can make a one-time donation, or you can set up a monthly contribution too.
 Our first goal is to sponsor our new Intern, and if we raise more than we need, we'll try to sponsor more than one.
 
-Even if Mixxx is free, please be aware that collecting money isn't.
+Even though Mixxx is free, please be aware that collecting money isn't.
+For now, the most efficient way to donate is with a direct bank transfer.
+You can click on your donation to see how the exact fees were calculated, but generally there will be 10% deducted to pay for our fundraiser host and bank handling.
+There are additional charges for credit card processing and, if applicable, foreign transactions.
+We have vetted these organizations and they providing a great service to open source projects like ours, so we are confident that this portion of the money is going to a good place.
```
ronso0:
```suggestion
We have vetted these organizations and they are providing a great service to open source projects like ours, so we are confident that this portion of the money is going to a good place.
```

pages/news/2020-09-23-outreachy.html
```diff
@@ -0,0 +1,57 @@
...
+That's where you come in!
+
+Our users have told us they want to give back, even though Mixxx is and always will be free.
+This is your chance to show your love for Mixxx.
+We've set up a fundraiser on OpenCollective so you can help Mixxx expand its contributor base and become a more vibrant, sustainable project.
+You can make a one-time donation, or you can set up a monthly contribution too.
+Our first goal is to sponsor our new Intern, and if we raise more than we need, we'll try to sponsor more than one.
+
+Even though Mixxx is free, please be aware that collecting money isn't.
+For now, the most efficient way to donate is with a direct bank transfer.
```
daschuer:
```suggestion
For now, the most efficient way to donate from the U.S. is with a direct bank transfer. For PayPal and credit card processing extra fees are charged from the donated money. 
```

pages/news/2020-09-23-outreachy.html
```diff
@@ -0,0 +1,57 @@
...
+You can click on your donation to see how the exact fees were calculated, but generally there will be 10% deducted to pay for our fundraiser host and bank handling.
+There are additional charges for Paypal or credit card processing and, if applicable, foreign transactions.
```
daschuer:
```suggestion
The only option for donaing from foreign countries is using PalyPal. PayPal silently adds 5% fee on top of the donated amount if currency conversion is required.   
```

daschuer (changes requested):

ywwg:
incorporated changes and fixed a couple typos

daschuer (approved):
LGTM, thank you.

ywwg:
Does our system have draft support? i.e. can we merge now and it will go live on the 1st?  Or do we need to wait

PR #153, opened 2020-10-07
update outreachy post

ywwg:
add gofundme links


daschuer:
Thank you. 
I wonder if it is obvious enough that one is in $ and the other in €? 
Maybe one has an idea to decorate the donation buttons a bit?

daschuer:
But anyway, LGTM. 

Will you also publish a social media post?

ywwg:
yes I'll do a repost -- good idea about the currency markers!

PR #155, opened 2020-10-16
Add various fixes and pre-commit support

Holzhaus:


Holzhaus:
Merge? This fixes some typos on the website as well.

ronso0:
LGTM, except I can't judge the yml/yaml files.

Holzhaus:
OK, let's merge then. Worst that can happen is that the hooks fail and we need to ignore them until they are fixed.

ronso0:
Alrigth.

PR #162, opened 2020-11-09
pages/features: Update controller and control vinyl images

Holzhaus:
The both the controller and control vinyl photos are very old and not
available for purchase anymore. This updates both photos.

The controllers in the images are the Pioneer DDJ-200, the Roland DJ-505
and the Denon MC7000 - all are supported by Mixxx 2.3.


pages/features.html
```diff
@@ -70,8 +70,8 @@ <h2><a class="headline-link" href="#hardware">{% trans "Customizable DJ Hardware
 
     <div class="grid-container halves u-align-left">
       <div>
-        <div style="max-height: 200px;">
-          <img class="responsive drop-shadow" src="{% static '/images/2.0/feature-midi.png' %}">
+        <div>
+          <img class="responsive drop-shadow" src="{% static '/images/2.3/feature-midi.png' %}">
```
Be-ing, line 74: Why is this file in a versioned directory? It's not a screenshot; it shouldn't change version to version.

Be-ing:
If we can put these images on the website, why not put controller images in the manual?

Holzhaus:
I'n not sure if these would cause issues as they are heavily edited, low resolution, overlayed with transparency/shadow effects. But you're right, better be on the safe side here.

Just for the record: Copyright was not the *only* argument I mentioned. Images without labels in the manual don't serve any useful purpose and unnecessary increase the file size.

Be-ing:
That was the opposite of my point. We have had the old image on the website for many years. No one has cared.

ronso0:
LGTM.
Images for the manual are an entirely different topic, and I agree with @Holzhaus here: if they are not labeled I don't see the purpose for adding them. Users looking into the manual for a specific controller already know how that looks like, so either add labeld greyscale SVGs or leave as it is.

Holzhaus:
Merge?

ronso0:
Oh, sorry, didn't mark this PR.

LGTM, thanks.

PR #167, opened 2020-12-12
Migrate to Pelican static site generator

Holzhaus:
Cactus is no longer maintained, depends on Python 2.7 which is EOL since January and it's not really suited for blogs.
This switches over to the [Pelican](https://github.com/getpelican/pelican) static site generator. It's works with recent Python 3 versions, it's actively maintained and has some neat features like filtering articles by tags, authors or date. I added tags to all existing articles, and also transformed some HTML posts to pure Markdown while I was fixing some Pelican warnings (missing alt attribute on img tags).
 
**TODO**
- [x] Fix video embeds
- [x] Make article metadata CSS responsive

**EDIT:** This mustn't be merged until we switched over mixxx.org to Netlify. Jenkins probably doesn't know that you now need to run `pelican` instead of `cactus build`.

PR #169, opened 2020-12-13
Netlify: Fix discourse comments blocked by CSP

Holzhaus:
Without this, discourse comments are blocked by our CSP.


Holzhaus:
ping, currently our discourse comment integration is broken.

I'll try to come up with a better solution once #167 has been merged.

Holzhaus:
That was quick :smile:

PR #172, opened 2020-12-15
Pelican fixes

Holzhaus:
Some minor improvements regarding author metadata. Also fixes the scons removal post, which disappeared because it's in the wrong directory now.


Holzhaus:
I'll just merge this myself to fix the broken article link and comment sections.

PR #174, opened 2020-12-15
theme: Prevent wrapping in navigation menu

Holzhaus:


Be-ing:
Thanks for fixing this. We should revisit #118...

PR #175, opened 2020-12-16
Fix macOS download link

Holzhaus:
Fixes https://mixxx.discourse.group/t/mac-download-does-not-work/20532


Be-ing:
If you're just fixing a broken link, I'm okay with pushing directly upstream without opening a PR.

PR #180, opened 2020-12-30
Improve download page

Holzhaus:
TODO:
- [x] Fix CSS on mobile
- [x] Add icons
- [x] ...

PR #177, opened 2020-12-28
Adding Outreachy internship blog to website.

deborahtrez:
I will be adding the blog posts about my internship experience with Outreachy  - while working to Improve the Mixxx DJ manual - to the Mixxx website.
This is the first PR (the first blog that I published).
@ywwg 
@Holzhaus

PR #181, opened 2020-12-30
Add full YAML metadata support

Holzhaus:
Prerequisite for #180.


Holzhaus:
I suggest to merge #177 before this, then give me time to add the necessary quotes to `title: ` strings (colons inside unquoted strings are syntax errors in YAML).

Holzhaus:
Alread merged as part of #180.

PR #187, opened 2021-01-10
Fix build warnings

Holzhaus:


Holzhaus:
Merging this myself now. We could have easily avoided that if we had merged #179.

PR #188, opened 2021-01-10
pages/download: Expand section if URL contains an anchor

Holzhaus:
This expands the appropriate section automatically when using a link like: https://mixxx.org/download/#stable-macos

### DIFF
--- a/theme/static/js/download.js
+++ b/theme/static/js/download.js
@@ -38,12 +38,21 @@
         return null;
     }
 
-    let os = detectOS();
-    if (os === null) {
-        return;
-    }
-
     document.addEventListener("DOMContentLoaded", function() {
+        // Expand section that has been navigated to with an anchor
+        if (document.location.hash) {
+            let anchorButton = document.querySelector(".accordion " + location.hash + " > input.expander");
+            if (anchorButton) {
+                anchorButton.checked = true;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


ronso0:
hmm.. it somehow "works" i.e. opens a container but the link https://mixxx.org/download/#stable-macos opens Ubuntu for me

ronso0:
..as does https://mixxx.org/download/#stable-windows

Holzhaus:
> hmm.. it somehow "works" i.e. opens a container but the link https://mixxx.org/download/#stable-macos opens Ubuntu for me

The code only runs when the page is loaded. Are you sure it really reloads the page? If you edit the URL fragment when you're already on the download site, the browser won't reload (press F5 to work around this).

ronso0:
Yes, opening https://mixxx.org/download/#stable-macos always opens the Ubuntu.
https://mixxx.org/download/#stable-windows too

ronso0:
/#fedora-stable works, though.
Seems it's just Windows and macOS that don't.

ronso0:
maybe because 'linux' user agent is detected?

Holzhaus:
> maybe because 'linux' user agent is detected?

Ah yes, that's true. Should be fixed now.

ronso0:
yup, works!

PR #197, opened 2021-01-19
Remove inline styles

Holzhaus:


Holzhaus:
Btw this also replaces the Mixxx logo in the splash and navbar with an SVG version because it looks "pixelated" in some situations, e. g. on my phone.

PR #216, opened 2021-03-14
explicitly say that Zulip is not for user support

Be-ing:


ronso0:
The change LGTM, but I doubt that those few people coming to Zulip for support do so because they found the link on the Get Involved page.
I rather think they pick it from the Community links at the bottom, and _I think_ they might miss the 'Forums' link because it's the only one in that column that lacks a fancy icon.
image

If someone joins Zulip, it may now seem a bit contradictory to find the top stream on the left is `_support`.
I don't remember what that was created for? More in-depth support with technical/programming issues?

Holzhaus:
We can improve it in another PR, let's merge :D

Be-ing:
> If someone joins Zulip, it may now seem a bit contradictory to find the top stream on the left is _support.

This was a hack before we switched the forum from phpBB to Discourse. People were struggling to get support because they were not getting the account activation emails from phpBB so they came to Zulip. I think we should delete the _support stream in Zulip now.

ronso0:
well, I guess it still contains valuable information. so if someone wants to look that up later it would be missing.
what would be the best way to deal with that?

Holzhaus:
Let's discuss here: https://mixxx.zulipchat.com/#narrow/stream/109695-_support/topic/delete.20this.20stream.3F

PR #217, opened 2021-03-18
A jack zero latency post

daschuer:


Be-ing:
build failed

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,46 @@
...
+
+Another disadvantage of Jack is that it can't deal (well) with two or more sound cards. Mixxx can do this using the ALSA API and this with also no extra latency as long the underlying driver allows it.
+
+## Conclusion 
+
+Don't use Jack when you don't need it for another reason.  The same will be true for Pipewire using the same architecture. 
+
+## What comes next? 
+
+There is a chance to get rid of the extra latency in these cases, and make Mixxx use the buffer provided by Jack/Pipewire. This requires that Mixxx becomes a native Jack or Pipewire application.
```
Be-ing: I don't understand why PortAudio can't do this.

daschuer: PortAudio Is only software, so it can be refactored to do this. And than Mixxx also need to be refactored to work with the new PortAudio API.

Be-ing: Why would this need a new PortAudio API? If PortAudio doesn't add any extra latency with ALSA then it is IMO a major bug if it does with JACK.

Be-ing: Have you reported this bug upstream to PortAudio?

Be-ing:
I'm not sure our blog is the best place to put this. IMO it belongs in a PortAudio bug report.

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: This is not a bug. It is an intended implementation detail. I had a discussion on the Portaudio Mailing list 5 years ago or such.

daschuer:
The intention is to get rid of the Jack has zero latency myth. 
I am under the impression you also get hooked on it. There is no need for regular Mixxx users to do the hassle of setting up Jack.

Be-ing:
We already have it documented in the manual to not use JACK if it is not required.

content/news/XXXX-XX-XX-jack-zero-latency.md
Be-ing: I cannot find the discussion in the PortAudio mailing list archive. If PortAudio does not add any latency with ALSA it should be possible to do the same with JACK.

Be-ing: If this bug is not fixable without a major refactoring, it may be better to write a new PipeWire backend for PortAudio.

daschuer: Yes, of cause, but that is not that easy, because Portaudio has an abstraction of sound cards. While Jack has just ports.
Portaudio gaps through Jack to get back the soundcard. This is nice to have a common interface for all APIs but comes at a cost of latency.

Be-ing: or maybe it would be better to switch to cpal after all...

Be-ing: I don't understand how PortAudio's abstractions require adding latency. If they don't for ALSA, they should not for JACK.

Be-ing:
> The intention is to get rid of the Jack has zero latency myth.

JACK does have zero latency. PortAudio adds latency. That is a bug.

daschuer:
>  I'm not sure our blog is the best place to put this. IMO it belongs in a PortAudio bug report.

This blog post not a bug report, just proves the current latency situation and give some insights how to measure round trip latency. 
The question if this is a bug or not is out of scope. But it gives an outlook that we have the chance to improve the situation. 

We have list of pending bugs here: 
https://bugs.launchpad.net/mixxx/+bugs?field.tag=jack

So hopefully this creates some momentum to improve the situation that Mixxx can start polished into the PipWire age.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
+title: "Has JACK zero latency?"
+authors: Daniel Schürmann
+tags: 2.3, jack
+date: 2021.03.18 02:42:18
+
+In regular intervals, we discuss how much latency the JACK layer introduces. Here we can read that Jack adds zero latency: [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html)
```
Holzhaus:
```suggestion
In regular intervals, we discuss how much latency the JACK layer introduces. The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
> There is **NO** extra latency caused by using JACK for audio input and output. When we say none, we mean absolutely zero.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
...
+This is true on its own, because Jack uses directly the buffer provided by ALSA to mix the audio sources together. ALSA has a second buffer that is used to feed the samples into the hardware. That's all.
```
Holzhaus:
```suggestion
This is true on its own, because JACK uses directly the buffer provided by ALSA to mix the audio sources together. ALSA has a second buffer that is used to feed the samples into the hardware. That's all.
```
And everywhere else

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
...
+
+However in the case of Mixxx another buffer is used, because Mixxx has its own real time mixing stage and uses the portaudio abstraction layer.
```
Holzhaus:
```suggestion
However in the case of Mixxx another buffer is used, because Mixxx has its own realtime mixing stage and uses the PortAudio abstraction layer.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
...
+
+When using the ALSA API directly Mixxx does what Jack does. It uses directly the ALSA buffer and adds this way zero latency.
```
Holzhaus:
```suggestion
When using the ALSA API directly Mixxx does what Jack does. It uses the ALSA buffer directly, which doesn't add any latency.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
...
+
+This can be confirmed by recording the own sound via a mic and measure the round trip latency:
+
+![Screenshot of audacity showing the round trip latency]({static}/images/news/roundtriplatency.png)
+
+
+The upper stream is the JACK case. The left channel is the recorded master 440 Hz sine wave and the right channel is the mic input.
+
+JACK is configured with a 1024 frames buffer and reports a latency of 46,4 ms for the sum of two buffers .
```
Holzhaus:
```suggestion
JACK is configured with a 1024 frames buffer and reports a latency of 46.4 ms for the sum of two buffers .
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,43 @@
...
+The roundtrip latency is 95 ms ~4 buffers = Driver + ALSA + 2 x Mixxx
+
+The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23,2 ms
```
Holzhaus:
```suggestion
The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23.2 ms
```

daschuer:
@Holzhaus thank you :-)

daschuer:
The batch commit button was grayed out for some reasons.

The file is prefixed with xxx but the date is set. How is this handled?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,44 @@
...
+
+The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23.2 ms
+The roundtrip latency is 49 ms ~2 buffers = Driver + ALSA
+
+Not in the picture is the ALSA pulse device. It runs at a latency of 104 ms ~5 buffers. Driver + ALSA + 2 x Pulse + ALSA
+
+With this picture we can verify that Mixxx actually has the same buffer size in both cases. When pressing pause. It fades this signal out over one buffer length these is equal in both cases.
+The peaks in the recorded right channel is the sound of the mouse click. You can only barely see the recorded sine wave.
+
+![Screenshot of audacity showing the fade out]({static}/images/news/fadeoutcompare.png)
```
Holzhaus, line 42: 2 minor nits:

- You could run audacity on the command line with LANG=C to make the user interface english
- I think audacity has a dark mode that you can use to make the screenshots fit better into the general theme of our website.

Be-ing:
https://github.com/PortAudio/portaudio/wiki/BufferingLatencyAndTimingImplementationGuidelines

:thinking:

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,44 @@
+title: "Has JACK zero latency?"
+authors: Daniel Schürmann
```
Holzhaus, line 2: Please add yourself to the `AUTHORS_METADATA` dict in the `pelicanconf.py`: https://github.com/mixxxdj/website/blob/website/pelicanconf.py#L158

Then run `tools/update_avatars.py`. It will automatically fetch your profile picture from github and add it to the source tree (don't forget to `git add` it).

Something like:

```python
AUTHOR_METADATA = {
    # ...
    "daschuer": {
        "name": "Daniel Schürmann", 
        "github": "daschuer",
        "discourse": "daschuer", 
        "tagline": "Mixxx Core Developer", 
    },
}
```

The author information including the avatar will be displayed below the post.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,44 @@
...
+tags: 2.3, jack
```
Holzhaus:
```suggestion
tags: 2.3, jack, alsa, sound
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,44 @@
...
+date: 2021.03.18 02:42:18
+
+In regular intervals, we discuss how much latency the JACK layer introduces. The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
```
Holzhaus: Let's give some more context to readers that aren't experts in the field. E. g.:
```suggestion
In regular intervals, we discuss how much latency the JACK Audio Connection Kit introduces. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux and it's a layer on top of the Advances Linux Sound Architecture (ALSA). The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,44 @@
+title: "Has JACK zero latency?"
+authors: Daniel Schürmann
+tags: 2.3, jack
+date: 2021.03.18 02:42:18
+
+In regular intervals, we discuss how much latency the JACK Audio Connection Kit introduces. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux and it's a layer on top of the Advances Linux Sound Architecture (ALSA). The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
```
Holzhaus: Fixing my own typo, sorry

```suggestion
In regular intervals, we discuss how much latency the JACK Audio Connection Kit introduces. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux and it's a layer on top of the Advanced Linux Sound Architecture (ALSA). The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
```

ishitatsuyuki:
I came from https://github.com/RustAudio/cpal/issues/560 and uh, this blog post does not explain anything.

> The question if this is a bug or not is out of scope.

It's not OK because it attributes what is apparently a PortAudio flaw to JACK's fault. There's a problem with how it is written. Also, it doesn't give enough information to make this actionable. Please at least link to the PortAudio mailing list discussion which you think is the cause.

daschuer:
Thank you @ishitatsuyuki, for looking into this.

> I came from RustAudio/cpal#560 and uh, this blog post does not explain anything.

This blog post was only intended to document the current situation and the way I have gathered the information. Do you recommend any changes?
I am happy to improve it. 

> It's not OK because it attributes what is apparently a PortAudio flaw to JACK's fault. 

Is it? It is written from the Mixxx perspective where the user hast the choice between Jack and ALSA API.

Please propose a change that this post becomes OK to you.

> Also, it doesn't give enough information to make this actionable. 

I have no yet a action to improve the situation. This is just a finding. And the acting from the user perspective is: Use ALSA API.

> Please at least link to the PortAudio mailing list discussion which you think is the cause.

I cannot find it after a quick search. I will try again.

I would be interesting to see the same graphs with CPAL. Can you share them at https://github.com/RustAudio/cpal/issues/560

May be we can use them for a follow up post do document the current state with CPAL.

ishitatsuyuki:
> This blog post was only intended to document the current situation and the way I have gathered the information. Do you recommend any changes?
I am happy to improve it.

~~You quote the JACK zero latency claim in the beginning of the post. This makes it look like you're attacking the claim.~~

Edit: After reading the post again, I think this one is actually fine.

The conclusion is also too generalized. You should emphasize that the problem is specific to Mixxx. I think you should write like "For now, we recommend using Mixxx with the ALSA backends even if you are running JACK. The same applies to PipeWire as Mixxx uses JACK protocol to connect to it."

Finally, you claim that "This requires that Mixxx becomes a native Jack or Pipewire application" without any investigation into the PortAudio internals. The claim is unprofessional at best, and dubious at worst.

> I would be interesting to see the same graphs with CPAL.

Your blog post doesn't describe the exact method used to take the recording. The software used is probably Audacity, but how did you record the output buffers (in particular when using ALSA directly)? Did you use a loopback?

If what we want to know is whether a library's abstraction adds latency, then I think a good idea is to reimplement something like the [jack_iodelay](https://github.com/jackaudio/tools/blob/master/iodelay.c) tool with that library. I'll try doing that and report the results.

Be-ing:
I agree that it's not fair to imply that JACK is the problem here when the problem is PortAudio.

Be-ing:
> May be we can use them for a follow up post do document the current state with CPAL.

This would not be appropriate for our blog. Our blog is for communicating with Mixxx users. It is not a discussion forum for considering the technical challenges Mixxx faces; that is what we have Zulip for.

daschuer:
Thank you @ishitatsuyuki, I will update the blog post with your suggestions.

> Your blog post doesn't describe the exact method used to take the recording.

I have played a 440 Hz sine wave with Mixxx on the left channel and the microphone input on the right channel. The feedback loop was closed via air. Than I use the recording feature of Mixxx to record the result. Audacity is used vor visualize the results and to compare the recordings. 

I don't think you can use jack_iodelay directly because it does not use CPAL but the approach in general will work.

Holzhaus:
> > May be we can use them for a follow up post do document the current state with CPAL.
> 
> This would not be appropriate for our blog. Our blog is for communicating with Mixxx users. It is not a discussion forum for considering the technical challenges Mixxx faces; that is what we have Zulip for.

I agree that it's not the right place for discussion, but I don't think that was @daschuer's intention. I think it's interesting for our users to follow the development progress and how we tackle obstacles and challenges. Your build infrastructure blog post is an example of that.

Be-ing:
> I have played a 440 Hz sine wave with Mixxx on the left channel and the microphone input on the right channel. The feedback loop was closed via air. Than I use the recording feature of Mixxx to record the result.

How did you do this considering Mixxx cannot record separate channels, only a single stereo mix?

Be-ing:
> I think it's interesting for our users to follow the development progress and how we tackle obstacles and challenges. Your build infrastructure blog post is an example of that.

Yes, I agree. But that post was made after we had a solution to show, not to start a discussion.

daschuer:
>>  I have played a 440 Hz sine wave with Mixxx on the left channel and the microphone input on the right channel. The feedback loop was closed via air. Than I use the recording feature of Mixxx to record the result.

> How did you do this considering Mixxx cannot record separate channels, only a single stereo mix?

Via the balance effect. I have added it to the blog.

daschuer:
>> May be we can use them for a follow up post do document the current state with CPAL.

> This would not be appropriate for our blog. Our blog is for communicating with Mixxx users. It is not a discussion forum for considering the technical challenges Mixxx faces; that is what we have Zulip for.

Yes right. May be we can reference the results as an amendment, to this if it contains user relevant Info.

daschuer:
@ishitatsuyuki I think I have addressed your notes. Thank you for the Input.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,57 @@
...
+  3114.842 frames     70.631 ms total roundtrip latency
+	extra loopback latency: 42 frames
+	use 21 for the backend arguments -I and -O Inv
+```
+
+The result is 70 ms ~3 buffers = Driver + ALSA + jack_iodelay. This is one buffer more than a native ALSA implementation.
+
+## Conclusion
+
+For now, we recommend using Mixxx with the ALSA backends even if you are running JACK. The same applies to [PipeWire](https://pipewire.org) as Mixxx uses JACK protocol to connect to it."
```
ishitatsuyuki: You copied over the quote.

ishitatsuyuki:
I think the 1 frame of delay probably comes from JACK trying to synchronize the input and output device. Does your input and output run on the same audio interface? If not, then the latency is inevitable due to JACK’s synchronized nature.

The one more frame of delay on Mixxx might be there due to a similar reason. Does Mixxx use PortAudio’s duplex abstraction? (I’m not sure if that is properly implemented though.) If not, how does it synchronize the input to the output?

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: Ups, I have forced push.

ishitatsuyuki (approved):

daschuer:
> Does your input and output run on the same audio interface? 

On the same audio interface. 

> If not, then the latency is inevitable due to JACK’s synchronized nature.

Last time I have checked how to use two soundcards with JACK, it was kind of broken with drop outs and such. Has this situation improved lately?  

> The one more frame of delay on Mixxx might be there due to a similar reason. Does Mixxx use PortAudio’s duplex abstraction? 

Yes, jack copies all streams into single buffers, and PortAudio undoes the work. In Mixxx we have a single callback for every sound hardware. See: https://github.com/mixxxdj/mixxx/blob/3d22a1caae325a8980d595e3c84abb0366fed261/src/soundio/sounddeviceportaudio.cpp#L42

> (I’m not sure if that is properly implemented though.) If not, how does it synchronize the input to the output?

I am not sure, but from the code in https://github.com/PortAudio/portaudio/blob/master/src/hostapi/jack/pa_jack.c it looks like PortAudio squeezes Jack a lot to make it look like a single sound card and re-use their own infrastructure.

Be-ing:
I cannot replicate these results. The top stereo track was with JACK, the bottom with ALSA. The right (bottom) channel in each was Mixxx's mic input. I used a loopback cable rather than a mic. I used a very big 2048 sample buffer to exaggerate the effect.
image

daschuer:
It is hard to see how the left channel relates to the right channel. It looks like the playback track is mixed in right and the left channel. Please verify the balance effect setting.

Can you redo the test using a 440 Hz track and start recording in the pre-roll?
This way you get rid of the fade in effect.

Which buffer size did you use?

Be-ing:
Sorry, I did not zoom in far enough before to see the difference. I repeated the experiment using the technique of starting from the preroll.
image
The problem is not PortAudio, it is jackd. jackd (JACK2) in synchronous mode has the same latency as ALSA. jackd in asynchronous mode adds an extra buffer.

Be-ing:
Here are ALSA, jackd sync, PipeWire, and jackd async compared. I used a 2048 sample buffer at 48 kHz (I had [trouble getting PipeWire to run at 44.1 kHz](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/938)). Round trip latency with ALSA and jackd in synchronous mode is 182 ms:
image

PipeWire adds 22 ms:
image

jackd in async mode adds 43 ms:
image

Be-ing:
I repeated the measurements with ALSA and PipeWire using a realistic buffer size for Mixxx, 256 samples at 48 kHz:
image

Round trip latency with ALSA is 34 ms.

image

PipeWire again added 22 ms of latency, so that additional latency appears unrelated to the buffer size. :thinking:

ishitatsuyuki:
PipeWire uses timer based scheduling and it has a manually tuned headroom for DAC transfer, etc, and I guess that causes a fixed latency. 22ms sounds unfortunate but it might be tunable (not sure). See https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/FAQ#pipewire-buffering-explained.

Edit: you can set this via `api.alsa.headroom`.

Be-ing:
I think we should close this PR. I will make a new PR to add information to the manual about PipeWire soon.

Be-ing:
Thank you for the link @ishitatsuyuki. I experimented with the PipeWire configuration. These tests were with a 256 sample buffer at 48 kHz. The top track was with ALSA, the middle track was with `api.alsa.disable-batch = true` in alsa-monitor.conf, and the bottom track was with `api.alsa.headroom = 0`. With `api.alsa.disable-batch = true`, the latency was 35 ms, about 1 ms more than ALSA. I am guessing the extra 1 ms is from the resampler mentioned in the [PipeWire FAQ](https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/FAQ#pipewire-buffering-explained).

image


With `api.alsa.headroom = 0` and `api.alsa.disable-batch = false`, there was the extra 21 ms of latency compared to PipeWire with `api.alsa.disable-batch = true`:

image

From the [PipeWire wiki](https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/Configuration):

> api.alsa.headroom
> USB devices are marked as batch devices in alsa, for batch devices we configure period-size/2 of extra headroom. For the default period size of 1024 for example, a batch device will be configured with 512 period-size and 512 headroom. If you disable batch, it will be period-size and no extra headroom.

I have not yet experimented to evaluate if disabling `api.alsa.disable-batch` has a significant impact on whether xruns occur.

Be-ing:
I opened an [issue](https://github.com/jackaudio/jackaudio.github.com/issues/138) for the JACK website to mention the extra period added by JACK2 async mode.

Be-ing:
PipeWire with both `api.alsa.headroom = 0` and `api.alsa.disable-batch = true` has identical latency to ALSA. Commenting out `api.alsa.headroom = 0` adds the 1 ms of latency.
image

Be-ing:
I opened an [issue](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/940) to improve the PipeWire documentation regarding latency. I am curious what the response will be.

daschuer:
> I think we should close this PR. I will make a new PR to add information to the manual about PipeWire soon.

This PR documents the current behavior of the released Mixxx version and gives a valid conclusion now. IMHO this is polishied for the realease and actually worked as intended to make trades like @Be-ing  and @ishitatsuyuki sensitive for this issue.

What is the reason to hold this off. Do you have suggestions to tweak it?

Be-ing:
The blog is not the place for documentation. The manual is.

ishitatsuyuki:
> IMHO this is polishied for the realease and actually worked as intended to make trades like @Be-ing and @ishitatsuyuki sensitive for this issue.

Uh, making us sensitive to this issue by putting out incomplete investigation? That sounds more suitable for a Zulip discussion rather than blog post. The analysis in this blog post is severely lacking and I think there's not much value in publishing it. (Leaving aside the fact that we couldn't reproduce your result, which is also bad since the claims can be voided that way.)

daschuer:
You cannot reproduce my results? Which one you cannot reproduce?

Can you reproduce the jack_iodelay results?

ishitatsuyuki:
I mean, Be tried to reproduce your results yesterday and he got results more or less contradicting your conclusion.

daschuer:
> Uh, making us sensitive to this issue by putting out incomplete investigation?

For my understanding I have adds all missing information from users point of view. What else is missing?

daschuer:
@Be-ing: Can you confirm my measures in default settings?

ishitatsuyuki:
You only did one small experiment but did not perform any in-depth analysis or validation of your reasoning. And as you can see Be-ing's results have shown that much of your reasoning might be wrong.

His results show that **Using Mixxx with JACK or PipeWire is OK**, just don't run JACK in asynchronous mode and also tweak PipeWire's headroom for low latency.

daschuer:
The blog states basically: When you select Jack, you increase your latency. This is is true. @Be-has proved this in his second Picture.

For my understanding you can configure jack in sync mode to get around it. 

I am happy to add tia finding.

How can sync mode be enabled? 

I am curious to read your result. The test is very easy to do if you have access to a laptop.

ishitatsuyuki:
> The blog states basically: When you select Jack, you increase your latency. This is is true.

The reasoning is flawed. If you run JACK in synchronous mode, then there's no increase in latency.

It's totally possible for a conscious user that they have configured synchronous mode for another workload, e.g. DAW. In that case they can use JACK in Mixxx without any drawbacks.

I initially gave approval to this PR despite the lack of investigations, but now I have changed my mind. I think it needs massive rework (it's not even re-work because there was little investigations) to be something that is interesting to the reader. In the current state, it's just a piece of experiment data waiting to be properly explained.

Also, I have a non pro-audio setup which is a much different situation from what you are testing. My microphone runs on USB from a different ADC with different sampling rate. It's likely that the test results would be irrelevant to this topic. So I'm not going to bother with testing right now.

daschuer:
> If you run JACK in synchronous mode, then there's no increase in latency.

How do I change the setting?

> In the current state, it's just a piece of experiment data waiting to be properly explained.

What else needs to be explained? 

> Also, I have a non pro-audio setup which is a much different situation from what you are testing. 

My setup is also far away from a pro setup. I use a fresh installed Ubuntu Focal 20.4 LTS and a general purpose laptop. Do you have access to a similar setup?

> My microphone runs on USB from a different ADC with different sampling rate. 

Is jack able to handle this without glitches?

Be-ing:
QJackCtl and Cadence both have checkboxes to toggle synchronous mode.

Be-ing:
> I mean, Be tried to reproduce your results yesterday and he got results more or less contradicting your conclusion.

To clarify, my results are consistent with @daschuer's results. But @daschuer's conclusion is misleading. The conclusion is:

> Using Mixxx with JACK or PipeWire is OK, just don't run JACK in asynchronous mode and also tweak PipeWire's headroom for low latency.

Again, I think this belongs in the manual, not the blog. I don't think our users care to see the data from these experiments. What matters is the conclusion.

However, I do think it would be helpful to make a blog post about using Mixxx with PipeWire on different distributions after we have updated the manual. In short, Fedora users should update to Fedora 34 when it is released in a few weeks. Arch users should install the PipeWire packages. Debian and Ubuntu users should stick with the setups they have been using until their distribution switches to PipeWire for audio (or consider switching to Fedora or Arch if they want the benefits of PipeWire sooner).

daschuer:
How can I switch between sync and asynchrone mode? Do I need to recompile Mixxx?

Be-ing:
Synchronous versus asynchronous mode is a runtime option of jackd.

Holzhaus:
> Again, I think this belongs in the manual, not the blog. I don't think our users care to see the data from these experiments. What matters is the conclusion.

I agree that we should add a hint box to the manual that JACK should be run in synchronous mode for better latency.

I do think this could be an interesting a blog post for users that want to test how much latency their setup has themselves.

daschuer:
> Synchronous versus asynchronous mode is a runtime option of jackd

How can I change that? All my attempts failed.

Holzhaus:
> QJackCtl and Cadence both have checkboxes to toggle synchronous mode.

daschuer:
I have exactly the same results on a Fedora 34 Live: 

Screenshot from 2021-03-21 06-26-29

This means every user will run into this. 

He will read "Jack introduces Zero latency" and will use Jack without verifying it. 

We have nothing about it in the manual and it is doubtful that users will read the manual again after we add this info. 
So a news post is IMHO the right place.

Be-ing:
> He will read "Jack introduces Zero latency" and will use Jack without verifying it.

Again, I've already opened issues upstream with both JACK and PipeWire to improve their documentation.

> We have nothing about it in the manual

I already said I would fix that.

> it is doubtful that users will read the manual again after we add this info.

It is highly unlikely users will dig back through old blog posts to find information about how to configure Mixxx. If users ignore the documentation that is at their own peril. Nothing has changed here. JACK has always worked this way.

Be-ing:
> I do think this could be an interesting a blog post for users that want to test how much latency their setup has themselves.

We already discuss how to measure round trip latency in the manual in the Using Microphones chapter. There is no other reason Mixxx users need to measure their round trip latency to use Mixxx. And linking to external resources for this is just a hack around Mixxx missing this functionality (Ardour has this built into its audio configuration preferences).

daschuer:
> QJackCtl and Cadence both have checkboxes to toggle synchronous mode.

I am probably bind or it is not there in my Version: 0.5.0

I have just found this on https://qjackctl.sourceforge.io: 

> 2021-03-14 - An End-of-Winter'21 Release.
> qjackctl-0.9.2.tar.gz

>  Clock source, Self-connect mode and Server synchronous mode options are now always enabled on the Setup / Settings dialog, although only available or effective to jackd2.

> 2018-11-23 - A Black-Friday'18 Release.
> Whether to use server synchronous mode option added to Setup / Settings / Parameters (only applied when JACK D-BUS interface is enabled).

daschuer:
The option is available in QJackCtl Version 0.9.1 Ubuntu Hirsute.

Be-ing:
You can use the `-S` option with jackd. I got confused before because the jackd man page is inconsistent with `jackd --help`. I have [reported this bug upstream](https://github.com/jackaudio/jack2/issues/736). The man page says:

> -S, --shorts
Try to configure card for 16-bit samples first, only trying 32-bits if unsuccessful.  (default: 32-bit samples)

`jackd --help` says:
> [ --sync OR -S ]

Be-ing:
We have found important gaps in upstreams' documentation through this discussion. That's great. We should update our own documentation and add links to upstreams' fixed documentation. But I don't think it's appropriate to post to our blog the technical details of how we discovered these shortcomings. I do think it would be a good idea to publish a blog post notifying users after we have updated the manual, but first we need to update the manual. I will write a blog post soon, but continuing to discuss why this PR shouldn't be merged is distracting from that.

daschuer:
Here are my results with Ubuntu Hirsute: 

Bildschirmfoto von 2021-03-21 13-10-11

* Top: Default Jack 
* Middle: Jack wit Jack D-Bus and Server Synchronous mode enabled, checkboxes in QJackCtl Version 0.9.1 
* Bottom: ALSA

The extra overall latency is due to the USB audio device in the virtual machine. 

So Jack sync mode is still not head up with ALSA.

What did you do to get rid of the remaining extra buffer?

daschuer:
> But I don't think it's appropriate to post to our blog the technical details of how we discovered these shortcomings. 

First you complain about not enough details, now the details are there and you are still complaining. 
I think this blog is in general a good thing and keeps our website interesting. 
I am also happy to receive detailed critics. But I am not fine with closing this one and wait until you write a better one. 

I am happy to have both.

Be-ing:
> USB audio device in the virtual machine

This introduces a ton of extra complexity. I am not going to spend time testing such an unrealistic use case or guessing how it might influence the results. If you want to test JACK2 in synchronous mode there is no need for a VM to do so.

Be-ing:
> I am also happy to receive detailed critics.

I am not going to do a thorough review of the text of this because it would be more work than writing an entirely new post that doesn't have a bunch of misconceptions.

daschuer:
> bunch of misconceptions.

Uh, that's hard. 

This is pointless critics without a bunch of comments.

Can you describe your setup that does not introduce extra latency? Was it original Jack or Pipewire? Which Distro did you use?

Be-ing:
I have no energy to repeat myself here.

daschuer:
I have integrated the findings about JACK sync mode. What other misconceptions are pending?

daschuer:
@Be-ing: Please share your setup that has configured Jack with the same latency than pure ALSA. Since this will probably consume some of the Mixxx processing time, I am curious to test it.

daschuer:
I have debugged through port audio and I have not found any extra latency. So we van blame port audio for it. 
This can be easily confirmed by some additional test with Mixxx and compared to the jack tools. 
I have described them here. Now all guessed facts are confirmed. 

I consider this now ready to merge.  Please review.

daschuer:
Can one have a look here? I think all issues have been fixed and it is well re-targeted to Linux users making the best of their latency. 
Is there anything left?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
+title: "Has JACK zero latency?"
+authors: Daniel Schürmann
+tags: 2.3, jack, alsa, sound, latency, pipewire
+date: 2021.03.18 02:42:18
```
Holzhaus:
```suggestion
status: draft
```
Our CI will update the date according to the date of the merge commit.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
+title: "Has JACK zero latency?"
```
Holzhaus: The JACK developers might misinterpret this as unfair criticism of JACK in general. What about something like:


```suggestion
title: "Does using Mixxx with JACK give you zero latency?"
```
Or something like that?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+authors: Daniel Schürmann
+tags: 2.3, jack, alsa, sound, latency, pipewire
```
Holzhaus:
```suggestion
tags: jack, alsa, sound, latency, pipewire
```
This is not 2.3-specific, right?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+date: 2021.03.18 02:42:18
+
+In regular intervals, we discuss how much latency the [JACK Audio Connection Kit](https://jackaudio.org) introduces when used in Mixxx. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux, and it's a layer on top of the [Advances Linux Sound Architecture (ALSA)](https://www.alsa-project.org). The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
+> There is **NO** extra latency caused by using JACK for audio input and output. When we say none, we mean absolutely zero.
+
+This is true on its own, because JACK uses directly the buffer configured for ALSA to mix the audio sources together. ALSA has a second buffer with the same length that is used to feed these samples into the hardware. That's all.
```
Holzhaus:
```suggestion
This is true on its own, because JACK uses the buffer configured for ALSA directly to mix the audio sources together. ALSA has a second buffer with the same length that is used to feed these samples into the hardware. That's all.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+However, in the case of Mixxx two more buffers are used by JACK, for syncing and to mixing the buffers of the patch field. It allows the clienet applications to use the full CPU time of each buffer cycle.
```
Holzhaus:
```suggestion
However, in the case of Mixxx two more buffers are used by JACK for syncing and mixing the buffers of the patch field. It allows the client applications to use the full CPU time of each buffer cycle.

This is not a problem of JACK per se, but of the way JACK support was implemented in Mixxx. 
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+![Screenshot of audacity showing the fade out]({static}/images/news/fadeoutcompare.png)
+
+For reference, I have done the same test using [jack_iodelay](http://manpages.ubuntu.com/manpages/bionic/man1/jack_iodelay.1.html)
+
+```
+  3114.842 frames     70.631 ms total roundtrip latency
+	extra loopback latency: 42 frames
+	use 21 for the backend arguments -I and -O Inv
+```
```
Holzhaus: Indent by 4 spaces instead of using fenced code blocks here. It improves readability in a text editor IMHO.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+date: 2021.03.18 02:42:18
+
+In regular intervals, we discuss how much latency the [JACK Audio Connection Kit](https://jackaudio.org) introduces when used in Mixxx. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux, and it's a layer on top of the [Advances Linux Sound Architecture (ALSA)](https://www.alsa-project.org). The [JACK FAQ](https://jackaudio.org/faq/no_extra_latency.html) state that:
+> There is **NO** extra latency caused by using JACK for audio input and output. When we say none, we mean absolutely zero.
+
+This is true on its own, because JACK uses directly the buffer configured for ALSA to mix the audio sources together. ALSA has a second buffer with the same length that is used to feed these samples into the hardware. That's all.
+
+However, in the case of Mixxx two more buffers are used by JACK, for syncing and to mixing the buffers of the patch field. It allows the clienet applications to use the full CPU time of each buffer cycle.
+
+When using the ALSA API directly Mixxx does what JACK does. It uses the ALSA buffer directly, which doesn't add any latency.
```
Holzhaus:
```suggestion
When using the ALSA API directly Mixxx does what JACK does: It uses the ALSA buffer directly, which doesn't add any latency. 
```

Be-ing (changes requested):
I still do not think this is appropriate content for the blog. Please, let's close this.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+This can be confirmed by recording the own sound via a mic and measure the round trip latency. For my test, I have used the internal speaker, microphone and ALC298 Analog sound. For separating the signals in one stereo stream, I have enabled the Balance effect for deck 1 fully right and for mic fully left. The test track is a 440 Hz sine wave generated by [Audacity](https://www.audacityteam.org). Then I have cued the track into the pre-roll start recording and press play.
```
Holzhaus:
```suggestion
This can be confirmed by recording the own sound via a mic and measure the round trip latency. For my test, I have used the internal speaker, microphone and ALC298 Analog sound. For separating the signals in one stereo stream, I have enabled the Balance effect to move deck 1 fully to the right channel and the mic fully to the left. The test track is a 440 Hz sine wave generated by [Audacity](https://www.audacityteam.org). Then I have cued the track into the pre-roll start recording and press play.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+The upper stream is the JACK case. The left channel is the recorded master 440 Hz sine wave and the right channel is the mic input.
+
+JACK is configured with a 1024 frames buffer and reports a latency of 46.4 ms for the sum of two buffers.
+The round trip latency is 95 ms ~4 buffers = Driver + ALSA + JACK Async mode + duplex stream
+
+The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23.2 ms
+The round trip latency is 49 ms ~2 buffers = Driver + ALSA
+
+Not in the picture is the ALSA pulse device. It runs at a latency of 104 ms ~5 buffers. Driver + ALSA + 2 x Pulse + ALSA
```
Holzhaus:
```suggestion
Not in the picture is the ALSA pulse device. It runs at a latency of 104 ms (driver + ALSA + 2 x pulse + ALSA = 5 buffers) 
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+This can be confirmed by recording the own sound via a mic and measure the round trip latency. For my test, I have used the internal speaker, microphone and ALC298 Analog sound. For separating the signals in one stereo stream, I have enabled the Balance effect for deck 1 fully right and for mic fully left. The test track is a 440 Hz sine wave generated by [Audacity](https://www.audacityteam.org). Then I have cued the track into the pre-roll start recording and press play.
+
+Here the resulting recordings visualized in [Audacity](https://www.audacityteam.org):
+
+![Screenshot of audacity showing the round trip latency]({static}/images/news/roundtriplatency.png)
+
+The upper stream is the JACK case. The left channel is the recorded master 440 Hz sine wave and the right channel is the mic input.
+
+JACK is configured with a 1024 frames buffer and reports a latency of 46.4 ms for the sum of two buffers.
+The round trip latency is 95 ms ~4 buffers = Driver + ALSA + JACK Async mode + duplex stream
```
Holzhaus:
```suggestion
The round trip latency is 95 ms (driver + ALSA + JACK async mode + duplex stream = 4 buffers). 
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23.2 ms
+The round trip latency is 49 ms ~2 buffers = Driver + ALSA
```
Holzhaus:
```suggestion
The round trip latency is 49 ms (driver + ALSA = 2 buffers). 
```

daschuer:
I am willing to make this appropriate. What needs to be changed?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+For reference, I have done the same test using [jack_iodelay](http://manpages.ubuntu.com/manpages/bionic/man1/jack_iodelay.1.html)
+
+```
+  3114.842 frames     70.631 ms total roundtrip latency
+	extra loopback latency: 42 frames
+	use 21 for the backend arguments -I and -O Inv
+```
+
+The result is 70 ms ~3 buffers = Driver + ALSA + JACK Async mode. The duplex cycle is omitted here. This is one buffer more than a native ALSA implementation.
```
Holzhaus:
```suggestion
The result is 70 ms (driver + ALSA + JACK async mode = 3 buffers). The duplex cycle is omitted here. This is one buffer more than a native ALSA implementation.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+To verify that the duplex cycle is not introduced in the Mixxx or [PortAudio](http://www.portaudio.com) abstraction layer used by Mixxx. I have routed the jack_iodelay signal trough Mixxx.
```
Holzhaus:
```suggestion
To verify that the duplex cycle is not introduced by Mixxx or the [PortAudio](http://www.portaudio.com) abstraction layer used by Mixxx, I have routed the `jack_iodelay` signal through Mixxx.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+Not in the picture is the ALSA pulse device. It runs at a latency of 104 ms ~5 buffers. Driver + ALSA + 2 x Pulse + ALSA
+PipeWire on Fedora 34 has by default the same latency as JACK in this test.
+
+With this picture we can verify that Mixxx actually has the same buffer size in both cases. When pressing pause, it fades the signal out over one buffer length which is equal in both cases.
+The peaks in the recorded right channel is the sound of the mouse click. You can only barely see the recorded sine wave.
+
+![Screenshot of audacity showing the fade out]({static}/images/news/fadeoutcompare.png)
+
+For reference, I have done the same test using [jack_iodelay](http://manpages.ubuntu.com/manpages/bionic/man1/jack_iodelay.1.html)
```
Holzhaus:
```suggestion
For reference, I have done the same test using [`jack_iodelay`](http://manpages.ubuntu.com/manpages/bionic/man1/jack_iodelay.1.html)
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+![Jack patch field showing the jack_iodelay Mixxx loop]({static}/images/news/jackpatch.png)
+
+
+```
+  2048.000 frames     46.440 ms total roundtrip latency
+	extra loopback latency: 2047 frames
+	use 1023 for the backend arguments -I and -O
+```
+
+This is the minimum we can expect, one buffer is needed for jack_iodelay and one for Mixxx. The result can be confirmed with the same setup using jack_latent_client which just passes the input to the output.
```
Holzhaus:
```suggestion
This is the minimum we can expect, one buffer is needed for `jack_iodelay` and one for Mixxx. The result can be confirmed with the same setup using `jack_latent_client` which just passes the input to the output.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+If we connect the output of Mixxx to the input, an extra buffer is used for some reasons. This can be also confirmed with jack_latent_client and the same feedback loop.
```
Holzhaus:
```suggestion
If we connect the output of Mixxx to the input, an extra buffer is used for some reason. This can be also confirmed with `jack_latent_client` and the same feedback loop.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+![Jack feedback loop]({static}/images/news/feedbackloop.png)
+
+The upper stream shows the Mixxx results and the lower stream shows the jack_latent_client results.
```
Holzhaus:
```suggestion
The upper stream shows the Mixxx results and the lower stream shows the `jack_latent_client` results.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+In addition, most distros use JACK with it's default "Server Asynchronous Mode" which introduce one buffer extra latency (--async-latenc) when accessing the sound card. From Ubuntu Hirsute 21.4, QJackCtl exposes a "Use server synchronous mode" checkbox in the sound card preferences. It is grayed out by default but becomes active if "Enable JACK D-Bus interface" is checked as well. In this case the JACK mixing is done after Mixxx in the same time interval.
```
Holzhaus:
```suggestion
In addition, most distros use JACK with it's default "Server Asynchronous Mode" which introduce one buffer extra latency (`--async-latency`) when accessing the sound card. From Ubuntu Hirsute 21.4, QJackCtl exposes a "Use server synchronous mode" checkbox in the sound card preferences. It is grayed out by default but becomes active if "Enable JACK D-Bus interface" is checked as well. In this case the JACK mixing is done after Mixxx in the same time interval.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Conclusion
+
+For now, we recommend using Mixxx with the ALSA backends even if you are running JACK. The same applies to [PipeWire](https://pipewire.org) as Mixxx uses JACK protocol to connect to it.
+
+It should be also noted that JACK can't deal (well) with two or more sound cards. Mixxx can do this using the ALSA API and this with no extra latency as long the underlying driver allows it.
+
+Pipewire will become default on most diestros and we need to find out what is the best setup for using it with Mixxx. Do you have interest to help? Get in contact with us at [Zulip](https://mixxx.zulipchat.com)
```
Holzhaus:
```suggestion
Pipewire will become default on most distros and we need to find out what is the best setup for using it with Mixxx. Do you have interest to help? Get in contact with us at [Zulip](https://mixxx.zulipchat.com)
```

Be-ing:
Nothing. The whole premise is inappropriate for the blog. Our blog isn't a scientific journal to publish results of experiments.

Holzhaus:
> Nothing. The whole premise is inappropriate for the blog. Our blog isn't a scientific journal to publish results of experiments.

I think it is of interest for our Linux users.

We offer different audio APIs, and users ask themselves: "what should I use?" JACK is considered the gold standard for pro audio on Linux (at least until Pipewire is more widely used). I do think it's interesting to point out that JACK isn't the best option for Mixxx if you aim for minimum latency. Even better, this post can be linked to if users want to measure the latency of their setup themselves.

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: Ah, yes.

Be-ing:
> We offer different audio APIs, and users ask themselves: "what should I use?"

That's what the manual is for. If we update the manual, I think it could be a good idea to announce that on the blog. But experimental data IMO does not belong on the blog. There is no point to proceed with this post. It would need a complete rewrite after updating the manual, so it will be easier to just write a new post.

> I do think it's interesting to point out that JACK isn't the best option for Mixxx if you aim for minimum latency.

This is just not correct.

Holzhaus:
@daschuer there is some trailing whitespace. Please fix.

Holzhaus:
> > We offer different audio APIs, and users ask themselves: "what should I use?"
> 
> That's what the manual is for.

Sure, we should mention it there. But it's not the place to explain how to measure latency. 

> If we update the manual, I think it could be a good idea to announce that on the blog. But experimental data IMO does not belong on the blog. There is no point to proceed with this post. It would need a complete rewrite after updating the manual, so it will be easier to just write a new post.

I disagree. This post is clearly Mixxx-related. I also believe it's interesting to our users (not all of them ofc, but the same goes for most other posts, e.g. the infrastructure Blog post). It's not just some experimental data, it also gives some background why there is latency. It could be improved, e. g. Why latency is even important when using Mixxx, but that's just nice to have, not mandatory IMHO. I guess most readers will already know that. 

> > I do think it's interesting to point out that JACK isn't the best option for Mixxx if you aim for minimum latency.
> 
> This is just not correct.

Because...? If you want to convince someone it's usually a good idea to give at least one reason why they are wrong. 

If you believe that this blog post is factually wrong, misleading and/or leaves out crucial information, then please say so and explain why.

Be-ing:
> If you believe that this blog post is factually wrong, misleading and/or leaves out crucial information, then please say so and explain why.

I already have. Please don't ignore that.

daschuer:
The pending comments are now done.
@Holzhaus: Is the file name correct? I am unsure. 

I have also added a paragraph explaining latency.

daschuer:
> I already have. Please don't ignore that.

This issue should be fixed now. Please verify.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,77 @@
...
+
+![Jack feedback loop]({static}/images/news/feedbackloop.png)
+
+The upper stream shows the Mixxx results and the lower stream shows the `jack_latent_client` results.
+
+In addition, most distros use JACK with it's default "Server Asynchronous Mode" which introduce one buffer extra latency (`--async-latency`) when accessing the sound card. From Ubuntu Hirsute 21.4, QJackCtl exposes a "Use server synchronous mode" checkbox in the sound card preferences. It is grayed out by default but becomes active if "Enable JACK D-Bus interface" is checked as well. In this case the JACK mixing is done after Mixxx in the same time interval.
+
+## Conclusion
+
+For now, we recommend using Mixxx with the ALSA backends even if you are running JACK. The same applies to [PipeWire](https://pipewire.org) as Mixxx uses JACK protocol to connect to it.
```
Holzhaus, line 74:
```suggestion
For now, we recommend using Mixxx with the ALSA backends even if you are running JACK. The same applies to [PipeWire](https://pipewire.org) as Mixxx uses JACK protocol to connect to it.
If you do want to use JACK, make sure to enable synchronous mode for better latency.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,77 @@
...
+
+It should be also noted that JACK can't deal (well) with two or more sound cards. Mixxx can do this using the ALSA API and this with no extra latency as long the underlying driver allows it.
```
Holzhaus: Is this actually a problem of JACK, or is it just a downside of our portaudio-based JACK implementation? If the latter, we should make that clear.

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,77 @@
+title: "Does using Mixxx with JACK give you zero latency?"
+authors: Daniel Schürmann
+tags: jack, alsa, sound, latency, pipewire
+status: draft
+
+In regular intervals, we discuss how much latency the [JACK Audio Connection Kit](https://jackaudio.org) introduces when used in Mixxx. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux, and it's a layer on top of the [Advances Linux Sound Architecture (ALSA)](https://www.alsa-project.org).
+
+Unlike analog devices, digital audio devices process audio in time slices of samples stored in memory buffers. These are passed through the various layers. It takes at least the time of one of such buffers, but can be longer depending on the nature of these layers. The resulting time is the latency of a digital audio device.
```
Holzhaus:
```suggestion
Unlike analog devices, digital audio devices process audio in time slices of samples stored in memory buffers. These are passed through the various layers. The latency of a digital audio device depends on the number and size of those buffers.
The aim is to minimize the latency, so that pressing a button or turning a knob in Mixxx works without noticable delay, which would make DJing by ear impossible.
```

Holzhaus (commented):
> @Holzhaus: Is the file name correct? I am unsure.

Yes, looks correct.

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: I would not explicit recommend it, because I think it is not default for some reasons. It is obvious that Jack needs some time to compose the final buffer and I assume that this time is removed from the time Mixxx can use. 

It might be the case that you can use in Async mode a smaller buffer, resulting to the same over all latency with a bigger buffer in sync mode, but more predictable. 

At least we need to collect some experience before recommending it.

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: This is an issue of Jack, which is missing a clock sync routine. Mixxx is not able to apply it's own clock sync code in this case. I have tested that when implement the Mixxx syncing.

daschuer: This this is only partially related, should we remove the sentence?

Be-ing:
Can we please stop talking about this and get 2.3 released already?

content/news/XXXX-XX-XX-jack-zero-latency.md
daschuer: I have added links for reference.

daschuer:
OK, thank you for your help, this post has become really good now. Is there anything left to do?

Holzhaus (approved):

Holzhaus:
@ronso0 @uklotzde @ywwg any thoughts on this blog post?

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,78 @@
+title: "Does using Mixxx with JACK give you zero latency?"
+authors: Daniel Schürmann
+tags: jack, alsa, sound, latency, pipewire
+status: draft
+
+In regular intervals, we discuss how much latency the [JACK Audio Connection Kit](https://jackaudio.org) introduces when used in Mixxx. That is one of the [Sound APIs](https://manual.mixxx.org/2.3/en/chapters/preferences.html#sound-api) that Mixxx supports on Linux, and it's a layer on top of the [Advances Linux Sound Architecture (ALSA)](https://www.alsa-project.org).
+
+Unlike analog devices, digital audio devices process audio in time slices of samples stored in memory buffers. These are passed through the various layers. The latency of a digital audio device depends on the number and size of those buffers.
+The aim is to minimize the latency, so that pressing a button or turning a knob in Mixxx works without noticeable delay, which would make DJing by ear impossible.
```
ronso0:
```suggestion
The aim is to minimize the latency, so that the audible result of pressing a button or turning a knob in Mixxx is produced without noticeable delay, in order to allow DJing by ear.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,78 @@
...
+
+This is true on its own, because JACK uses the buffer configured for ALSA directly to mix the audio sources together. ALSA has a second buffer with the same length that is used to feed these samples into the hardware. That's all.
+
+However, in the case of Mixxx two more buffers are used by JACK for syncing and mixing the buffers of the patch field. It allows the client applications to use the full CPU time of each buffer cycle.
+
+This is not a problem of JACK per se, but of the way JACK support was implemented in Mixxx.
+
+When using the ALSA API directly Mixxx does what JACK does: It uses the ALSA buffer directly, which doesn't add any latency.
+
+This can be confirmed by recording the own sound via a mic and measure the round trip latency. For my test, I have used the internal speaker, microphone and ALC298 Analog sound. For separating the signals in one stereo stream, I have enabled the Balance effect to move deck 1 fully to the right channel and the mic fully to the left. The test track is a 440 Hz sine wave generated by [Audacity](https://www.audacityteam.org). Then I have cued the track into the pre-roll start recording and press play.
```
ronso0:
```suggestion
This can be confirmed by recording the own sound via a mic and measure the round trip latency. For my test, I have used the internal speaker, microphone and ALC298 Analog sound. For separating the signals in one stereo stream, I have enabled the Balance effect to move deck 1 fully to the right channel and the mic fully to the left. The test track is a 440 Hz sine wave generated by [Audacity](https://www.audacityteam.org). Then I have cued the track into the pre-roll, started recording and pressed play.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,78 @@
...
+
+Here the resulting recordings visualized in [Audacity](https://www.audacityteam.org):
```
ronso0:
```suggestion
These are the resulting recordings visualized in [Audacity](https://www.audacityteam.org):
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,78 @@
...
+
+![Screenshot of audacity showing the round trip latency]({static}/images/news/roundtriplatency.png)
+
+The upper stream is the JACK case. The left channel is the recorded master 440 Hz sine wave and the right channel is the mic input.
+
+JACK is configured with a 1024 frames buffer and reports a latency of 46.4 ms for the sum of two buffers.
+The round trip latency is 95 ms (driver + ALSA + JACK async mode + duplex stream = 4 buffers).
+
+The lower stream is the ALSA case. Mixxx is configures with the same single buffer of 1024 frames = 23.2 ms
```
ronso0:
```suggestion
The lower stream is the ALSA case. Mixxx is configured with the same single buffer of 1024 frames = 23.2 ms.
```

content/news/XXXX-XX-XX-jack-zero-latency.md
```diff
@@ -0,0 +1,78 @@
...
+
+This is the minimum we can expect, one buffer is needed for `jack_iodelay` and one for Mixxx. The result can be confirmed with the same setup using `jack_latent_client` which just passes the input to the output.
+
+If we connect the output of Mixxx to the input, an extra buffer is used for some reason. This can be also confirmed with `jack_latent_client` and the same feedback loop.
+
+![Jack feedback loop]({static}/images/news/feedbackloop.png)
+
+The upper stream shows the Mixxx results and the lower stream shows the `jack_latent_client` results.
+
+In addition, most distros use JACK with it's default "Server Asynchronous Mode" which introduce one buffer extra latency (`--async-latency`) when accessing the sound card. From Ubuntu Hirsute 21.4, QJackCtl exposes a "Use server synchronous mode" checkbox in the sound card preferences. It is grayed out by default but becomes active if "Enable JACK D-Bus interface" is checked as well. In this case the JACK mixing is done after Mixxx in the same time interval.
```
ronso0:
```suggestion
In addition, most distros use JACK with its default "Server Asynchronous Mode" which introduce one buffer of extra latency (`--async-latency`) when accessing the sound card. In Ubuntu Hirsute 21.4, QJackCtl exposes a "Use server synchronous mode" checkbox in the sound card preferences. It is grayed out by default but becomes active if "Enable JACK D-Bus interface" is checked as well. In this case the JACK mixing is done after Mixxx in the same time interval.
```

ronso0 (changes requested):
Nope, none. Never bothered with Jack, though I do understand the post and fixed some typos.

Holzhaus:
> Nope, none. Never bothered with Jack, though I do understand the post and fixed some typos.

But do you think it's interesting for our users? Or do think this doesn't belong on our website?

ronso0:
Dunno.. I guess if someone considers switching to Jack for a more advanced setup this is probably interesting. While the same person might skip the DVS timecode post. It's all relevant for Mixxx so it should go to the blog, sure.

Holzhaus:
Okay, I think we have sufficient backing by mixxx devs that this can be interesting to our users and be published on the website (ofc our user base is diverse, so some users won't care - like any blog post that isn't a release announcement).

@daschuer please fix the typos that @ronso0 pointed out, then we can merge.

daschuer:
```
WARNING: Missing avatar for author 'Daniel Schürmann', expected /home/runner/work/website/website/content/images/avatars/daschuer.png
ERROR: Could not process news/XXXX-XX-XX-jack-zero-latency.md
  | Warning encountered
CRITICAL: Error encountered
Error: Process completed with exit code 1.
```

CI is failing. What need to be done to remove that warning?

Holzhaus:
Run `tools/update_avatars.py` and commit the png file.

daschuer:
The warning complains about the wrong path:
/home/runner/work/website/website/content/images/avatars/daschuer.png
/home/runner/work/website/website/theme/static/images/avatars/daschuer.png

Holzhaus:
Please rebase on latest website, maybe it's already been fixed.

Holzhaus:
> Please rebase on latest website, maybe it's already been fixed.

Nope, the update_avatars,py script is just out of date. I'll remove it, because it's obsolete.

You can just build the page locally and it automatically fetches the avatar to the right directory:

```
(venv) $ pelican
WARNING: Feeds generated without SITEURL set properly may not be valid
WARNING: Missing avatar for author 'Daniel Schürmann', expected /home/jan/Projects/mixxx-website/content/images/avatars/daschuer.png
WARNING: Downloaded missing avatar to: /home/jan/Projects/mixxx-website/content/images/avatars/daschuer.png
Done: Processed 140 articles, 1 draft, 9 pages, 0 hidden pages and 0 draft pages in 3.06 seconds.
(venv) $ git status
On branch test
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	content/images/avatars/daschuer.png

nothing added to commit but untracked files present (use "git add" to track)
```

daschuer:
The issue persists after rebase.

Holzhaus:
> The issue persists after rebase.

Sorry, that comment was wrong, try this please: https://github.com/mixxxdj/website/pull/217#issuecomment-835578888

Holzhaus:
> Nope, the update_avatars,py script is just out of date. I'll remove it, because it's obsolete.

https://github.com/mixxxdj/website/pull/240

Holzhaus:
In short, run these commands:

```bash
$ python3 -m venv venv                    # Create a virtual environment named "venv"
$ source venv/bin/activate                # Activate the virtual environment
(venv) $ pip install -r requirements.txt  # Install requirements to build website into the venv
(venv) $ pelican                          # Build website
(venv) $ deactivate                       # Leave virtual environment
$ git add content/images/avatars
```

Next time you want to build the website:

```bash
$ source venv/bin/activate                # Activate the virtual environment
(venv) $ pelican                          # Build website
```

Holzhaus:
Thank you!

PR #222, opened 2021-03-25
Update development cycle image

Swiftb0y:
I kept the edits as minimally invasive as possible since this is not my area of expertise and I didn't want to break anything.


Holzhaus (approved):
Thanks!

PR #223, opened 2021-03-26
Add support for fetching package metadata

Holzhaus:
Testing instructions:
1. Download https://gist.github.com/cd8e39e13cdccd3e8f7d6c3556568ed8 and put in into some directory (e.g. `/tmp/foo`).
2. `cd` into that directory and run `python -m http.server 7999`
3. Apply this diff and rebuild locally:

```diff
diff --git a/content/pages/download.md b/content/pages/download.md
index 495d906..0013066 100644
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -6,6 +6,7 @@ versions:
     name: 2.3.0 Beta
     next_stable_version: 2.3.0
     release_announcement: /news/2020-06-07-mixxx-2-3-beta-released/
+    download_manifest: http://localhost:7999/manifest.json
     downloads:
       - slug: windows
         os: Windows 7 or later
```

This is how it should look:

Screenshot from 2021-03-26 15-04-50


daschuer:
How will this work? 
Will the Mixxx binary access this information?
Are we sure we can maintain this after mayor changes to our website?

Holzhaus:
> How will this work?

It will download a `manifest.json` file with the metadata from downloads.mixxx.org. I'll make a PR with the necessary changes to our mixxx repo CI soon-ish. It's currently WIP: https://github.com/mixxxdj/mixxx/pull/3747

> Will the Mixxx binary access this information?

No, but we could use this informstion eventually.

> Are we sure we can maintain this after mayor changes to our website?

I think so. Right now this doesn't change anything unless you apply the patch above.

Holzhaus:
Added a commit to use the manifest from the Deploy-metadata branch of the Mixxx repo. Please have a look at the Deploy Preview. If it looks okay, I'll remove the commit and we can merge.

daschuer:
Thank you, this looks good to me so far. 

I think this this can make use of a comment block that explains the feature. I think else it is hard to maintain the code later, if something is broken for some reasons. 

What is the reason for have the redundant sha256_url in the json file?

Holzhaus:
> I think this this can make use of a comment block that explains the feature. I think else it is hard to maintain the code later, if something is broken for some reasons.

I'll add some comments. 


> What is the reason for have the redundant sha256_url in the json file?

It lets users download the sha256sum file in case they want to verify integrity locally.

Holzhaus:
> > I think this this can make use of a comment block that explains the feature. I think else it is hard to maintain the code later, if something is broken for some reasons.
> 
> I'll add some comments.

Done.

daschuer:
This is still a draft, what else is missing? 

I have one general concern, because we have no longer a straight line of info for the user.  

```
mixxx git -> download.mixxx.org -> netlify -> html webside 
                         ^                      ^                              
                         |                      |
                          --------------------- user 
```
If something fails in the netifly stage, the user sees inconsistent info. 

The alternative would be to extract the json file on the fly. I can't explain how it works, but 
I guess there is a way to show json content inside a cactus page. 
Do you know a way?

Holzhaus:
> This is still a draft, what else is missing? 

I need to remove https://github.com/mixxxdj/website/commit/4c2b3be246db0f67310df4407e573a3a9827830a
before merging this. But without that commit we won't see anything, that's why I added it. 

> I have one general concern, because we have no longer a straight line of info for the user.  
> 
> ```
> mixxx git -> download.mixxx.org -> netlify -> html webside 
>                          ^                      ^                              
>                          |                      |
>                           --------------------- user 
> ```
> If something fails in the netifly stage, the user sees inconsistent info. 
> 
> The alternative would be to extract the json file on the fly. I can't explain how it works, but 
> I guess there is a way to show json content inside a cactus page. 
> Do you know a way?  

Using JavaScript. That would break it for all users that block JS via NoScript or use a terminal browser. 

But the information won't be inconsistent  anyway, just outdated. 

> If something fails in the netifly stage, the user sees inconsistent info. 

If the Netlify build fails, the website  Deploy won't finish and the website won't be updated. That means that the user will be presented with the old download (from the previous deploy) link and it's corresponding metadata.

To clarify: we currently have a download URL like `Mixxx-2.3-latest.msi`. When the manifest data is parsed, that is URL is replaced with the URL from the manifest (that also has the benefit that the file in the Download directory has a meaningful file name including the commit ID). 
That means that the metadata will never be inconsistent with the download URL unless the manifest.json contains inconsistent data.

Holzhaus:
> > This is still a draft, what else is missing? 
> 
> I need to remove https://github.com/mixxxdj/website/commit/4c2b3be246db0f67310df4407e573a3a9827830a
> before merging this. But without that commit we won't see anything, that's why I added it.

When https://github.com/mixxxdj/mixxx/pull/3748 has been merged I can add the actual manifest.json URL.

daschuer:
Ah OK sounds good. Can we copy that info somewhere to the code?

Holzhaus:
Btw, I updated the PR. As you can see the build-check fails because the URL doesn't exist yet (the deploy preview works because we do allow warnings for deploy previews, actual deployment would fail because of the warning). If you now check the deploy preview, it just shows the download page with the hardcoded values from download.md (without any metadata).

I'll force-push again to when the CI job on the mixxx repo finished to trigger a rebuild.

daschuer:
> WARNING: Failed to retrieve manifest URL: https://downloads.mixxx.org/builds/2.3/manifest.json

nice

Holzhaus:
Hmm, something is wrong. https://downloads.mixxx.org/builds/2.3/manifest.json is empty. Anyway, this needs to be fixed on the Mixxx repo. This PR is ready now.

Holzhaus:
> Hmm, something is wrong. https://downloads.mixxx.org/builds/2.3/manifest.json is empty. Anyway, this needs to be fixed on the Mixxx repo. This PR is ready now.

Ah, I accidently removed the `id` from the "Generate Artifact Metadata" step in https://github.com/mixxxdj/mixxx/pull/3748/commits/c71e64d211243e121b89fe19495576fbf58892cb which resulted in the job output being empty. I added it again in https://github.com/mixxxdj/mixxx/commit/3f26160d086f3f2f3cbffb6c673003618989bc0f. As soon as https://github.com/mixxxdj/mixxx/actions/runs/696026878 finished it should work.

Holzhaus:
Deploy Preview looks fine, ready to merge.

daschuer (approved):
Thank you very much for this feature.

PR #231, opened 2021-04-15
Download page improvements

Holzhaus:
- Renamed some package properties to match metadata from `manifest.json` (easier to read and less confusing).
- Renamed `windows32`/`windows64` slugs to `win32`/`win64`
- Removed default link (`Mixxx-branch-latest.*`, will become deprecated mixxxdj/mixxx#3789). If the manifest is not found or broken, it will no display a message that the download is currently unavailable.


Be-ing:
The Netlify preview shows that the download is not available. Is that because it is just a preview and not deployed to mixxx.org?

Holzhaus:
No, it's because I renamed some metadata members in the manifest and the current json file still uses the old names. That's why it's still a draft ;-)

Holzhaus:
The last commit changes the manifest to the one used in mixxxdj/mixxx#3791.  Please check the Deploy preview. I'll remove it before merge.

daschuer (approved):
LGTM

Holzhaus:
I removed the WIP commit. If you think that this and mixxxdj/mixxx#3791 are ready to merge, then please merge this one first, then mixxxdj/mixxx#3791.

PR #234, opened 2021-04-23
Fix Facebook URL

coliff:
Facebook links always have www


Holzhaus (approved):
Doesn't really make a difference, because Facebook makes a redirect to the www subdomain. But let's merge this anyway.

coliff:
yep, saves a HTTP redirect so is slightly faster :-)

PR #237, opened 2021-04-23
Generate download.json for update checker

Holzhaus:
Related Zulip Discussion: https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Update.20Notifier

### DIFF
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -4,6 +4,7 @@ slug: download
 template: pages/download
 versions:
   unstable:
+    version: 2.3
     name: 2.3.0 Beta
     next_stable_version: 2.3.0
     release_announcement: /news/2020-06-07-mixxx-2-3-beta-released/
@@ -75,7 +76,7 @@ versions:
           name: 2.3 branch
           file_url: https://github.com/mixxxdj/mixxx/archive/2.3.zip
   stable:
-    name: 2.2.4
+    version: 2.2.4
     release_announcement: /news/2020-06-25-mixxx-2-2-4-released/
     downloads:
       - slug: windows
@@ -140,6 +141,7 @@ versions:
           name: 2.2.4 release
           file_url: https://github.com/mixxxdj/mixxx/archive/release-2.2.4.tar.gz
   testing:
+    version: 2.4
     title: Development Snapshots
     text: |
       A great way to contribute to Mixxx is testing the latest code we're working on and giving early feedback. Refer to the [Testing wiki page](https://github.com/mixxxdj/mixxx/wiki/Testing) for where to find the latest builds and instructions how to test pull requests before they are merged.
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -1,5 +1,6 @@
 ---
 title: Download
+slug: download
 template: pages/download
 versions:
   unstable:
--- a/plugins/download_metadata/download_metadata.py
+++ b/plugins/download_metadata/download_metadata.py
@@ -45,6 +45,9 @@ def page_generator_context(page_generator, metadata):
     datetime_format = f"{date_format} %H:%M"
 
     for version_name, version_data in metadata.get("versions", {}).items():
+        if "name" not in version_data:
+            version_data["name"] = version_data["version"]
+
         # Check if a manifest URL is specified for this version and download it
         manifest_url = version_data.get("download_manifest")
         if not manifest_url:
--- a/plugins/download_metadata/download_metadata.py
+++ b/plugins/download_metadata/download_metadata.py
@@ -1,8 +1,18 @@
+import datetime
 import json
 import logging
-import datetime
+import os
+import os.path
 import urllib.request
 from pelican import signals
+
+
+class JSONEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, datetime.datetime):
+            return obj.isoformat()
+        else:
+            return json.JSONEncoder.default(self, obj)
 
 
 def format_size(num, suffix="B"):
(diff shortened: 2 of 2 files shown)


daschuer:
Thank you. This will be a great improvement to keep our users updated. However I am a bit lost how this will work. 
For my understanding we need a fixed URL with a file that can be read from any Mixxx version.

Will we have three streams just like from our ppa? 

This should be independent from any technology and hosting service. 
From that point of view I am am unsure if creating it with Netifly is the best approach. 

Can't it be a https://downloads.mixxx.org/manifest.json that just links the other manifests deeper in the directories. 

For the old releases we have a Todo list of manual actions in our wiki. One point will than be to update that file to switch the release stream. If that is done in turn with updating the website it is fine. I am just afraid that the list will still have some topics and unsure if automation of all is worth the work.

Holzhaus:
> For my understanding we need a fixed URL with a file that can be read from any Mixxx version.

Yes.
> Will we have three streams just like from our ppa?

Yes.

> This should be independent from any technology and hosting service.
> From that point of view I am am unsure if creating it with Netifly is the best approach.

It's built on netlify, but we can safely switch to another hoster if we want. The domain is independent. The file lives at `https://mixxx.org/download.json`. It can be hosted on any hoster that supports static sites. It's automatically built by pelican ever time you regenerate the website.

> Can't it be a https://downloads.mixxx.org/manifest.json that just links the other manifests deeper in the directories.

No, these manifest don't contain information about the release announcement and cannot be overridden manually if we ever need it (someone pushed a broken tag, or we want to change the manifest file format, or whatever).

Please have a look at the deploy preview, append `download.json` to the URL: https://deploy-preview-237--mixxx-website.netlify.app/download.json

Let me know what you think.

> For the old releases we have a Todo list of manual actions in our wiki. One point will than be to update that file to switch the release stream. If that is done in turn with updating the website it is fine. I am just afraid that the list will still have some topics and unsure if automation of all is worth the work.

Well, this just makes sure that the information on the website is in sync with the manifest. It would actually be more work to edit such a file manually.

daschuer:
Ah I understand. So this is "just" a machine readable table of content for our download page. This does not jet allow to generate a pop up message when a new version is available. 

What shall be the user experience? 

Something like this https://sparkle-project.org/:
grafik

There is also a windows port: https://github.com/vslavik/winsparkle including a Qt example. 

The related metadata file looks like that: 
https://winsparkle.org/example/appcast.xml


I think just using Sparkles directly will release us form some headaches about security concerns. I have good experience with it. But I am also open to implement our own solution.

Holzhaus:
> I think just using Sparkles directly will release us form some headaches about security concerns.

I think the opposite is the case tbh. What I had in mind was just a very simple update checker: retrieve metadata from the URL, if the version is higher open a popup notifiing the user with a link to the download page and the release announcement.

Even if that is hacked somehow, the users still need to download and run the executable themselves. So there won't be a way to do remote code execution or something. 

Integrating a full-blown updater will make it necessary to be extra careful with this feature. Also, this would increase the complexity and we'd have multiple different code paths for each OS and require more testing. I also see a DSA signature in the rss feed, and we'd need to worry how to automate generating that in a secure way. And we'd need to bring our release notes in a machine readable format.

We also need to implement OS/arch detection on the client and match that with the available packages on the server.

It will improve the user experience slightly, but I think the risks are higher and it needs much more work. If someone wants to work on that go ahead, but I planned something much simpler because it's less work and low risk.

daschuer:
The issue is that we have not much of a second chance. Once we have a check released in the Mixxx binary, we need to make it work for some years.

daschuer:
> Integrating a full-blown updater will make it necessary to be extra careful with this feature. 

Yes, that's why I trust an established framework more than any from our own solutions.

The verification with sparkles is done by providing the public key with mixxx and have the private key on our build server.

> And we'd need to bring our release notes in a machine readable format.

Sparkles is using an rss feed, which should be Ok, I think.

If we are concerned about the extra work, we should postpone the whole idea. For my feeling, 2.3.0 is just around the corner and there is a risk of doing it wrong, and no chance to fix it later.

daschuer:
@Holzhaus Would you mind to have a brief look into Sparkle? If there is no red flag that prevent us from using it, I can offer to help integrate it.

daschuer:
This seems to be the commit that integrates Sparkle into wireshark: 
https://github.com/wireshark/wireshark/commit/02057200fd73fc70cc4a6eeece5c033c14f0c50a

ronso0:
fwiw I'm also very much for the simple update notifier.
Just a super-simple-to-implement popup only, probably with a link to mixxx.org/download.
Users will end up at the download page, can read the release announcement, maybe also upgrade instructions per OS, list of known bugs, maybe check out the support links etc. etc.

For me the main benefit of the manual download is that you can take extra steps like backing up the settings and stuff IF I want and WHEN I want. With automatic updates I always get a little nervous because that's like taking the hands off the steering wheel downtown, but maybe that's just me and the aftermath of using Windows for 20years.

> If we are concerned about the extra work, we should postpone the whole idea. For my feeling, 2.3.0 is just around the corner and there is a risk of doing it wrong, and no chance to fix it later.

@Holzhaus
Did you plan to squeeze the notifier into 2.3??

daschuer:
> fwiw I'm also very much for the simple update notifier.

By default, Spakle gives you all the options. Would this work for you, or is it too scary? It can be configured to a different behavior if we like or become a opt-in preferences option. 

The default is head up with Ubuntu users using the PPA.

ronso0:
what scares me is having a external tool with tons of features where a lot can go wrong.
my Mixxxperience with all sorts of cross-platform features tells me that something WILL go wrong, which would then Mixxx be blamed for.
otoh if we disable all fancy features and auto-update there's no reason to use it instead of 

fetch download.json
show a simple popup with a hyperlink.  (except this lacks a "Remind me later" button :\

(this is me thinking the fetch + popup is easy to implement / maintain and absolutely riskfree for the rest of Mixxx)

daschuer:
I share your concerns, but I have the opposite conclusion. I consider maintaining our own update checker more work and a higher risk of having it not right initially. 

Sure am biased, because my Sparkle experience. 

We should at least have a look, before doing the work our self. 

I suggest to test Wireshark on Windows to see Sparkle in action:
grafik

Holzhaus:
FWIW this is my custom update checker that I originally planned: https://github.com/mixxxdj/mixxx/compare/2.3...Holzhaus:update-check
(the dialog is still missing and the version comparison is missing, everything else is done). I stopped working on it because you mentioned another solution might be preferred.

My main gripe with Sparkle is that it's not cross platform. Therefore we'll have multiple code paths and I cannot test this on Linux. Also, I also wonder how we the XML looks for multiple OSes, considering that this winsparkle is a port, not part of the main project.

But I'm open to it if you want to take care of it. I can take care of the XML Generation if you give me an example how it works for both platforms.

daschuer:
> My main gripe with Sparkle is that it's not cross platform.

As far as I have understand it, it is intentionally not cross platform, using the native API to offer the best user experience. The external interface is unified. 

I think there are pro and cons. On the cons site, we have less control and it might be more complex than is strictly required for us. On the pro site we have an established and secure solution that is done. Your one is now partially done. 

I am happy to jump in and do the work of integrating Sparkle. But I have no interest to do it against your conviction. 

Did you had a chance to test Sparkle in action? 

What do we need:
* put the sparkles library to our ci 
  * the best via vcpck. 
  * does it support cmake source build only? Unfortunately only the VC project is maintained. 
* Release announcements as rss feed or html. 
  * How does this play with Netifly/GitHub. 
  * Can we just link https://github.com/mixxxdj/mixxx/releases/tag/release-2.2.4  
* Signing. 
  * This should be not hard as we already sign the PPA builds. 
* Define a fixed URL location for the different "Streams" and Targets      
  * Wireshark uses the complex but mature: https://wiki.mozilla.org/Software_Update:Checking_For_Updates
  * https://www.wireshark.org/update/0/Wireshark/3.4.5/Windows/x86-64/en-US/stable.xml
  * https://www.wireshark.org/update/relnotes/wireshark-3.4.5.html

daschuer:
The Mac version is on homebrew :-) https://formulae.brew.sh/cask/sparkle

Swiftb0y:
I have to agree with @Holzhaus here. I understand that re-using existing solutions is usually desirable, but in this case it would using a sledgehammer to crack a walnut.
I'm mostly afraid that integration sparkles would be maintainance nightmare, in code, in building and packaging while most of the beneficial features you mentioned are not even really what we want. 
In addition, it seems like the majority of the work needed in mixxx and the server is already done.

daschuer:
I am not against building our own updater if we dislike Sparkle. 

> I'm mostly afraid that integration sparkles would be maintenance nightmare

From my experience you get a rock solid solution for the same maintenance or even less maintenance. The cost is less flexibility, we don't have anyway because once a version is out we need to maintain it for long.

For my understanding, nowadays user expect an auto update like on Linux even if they are on windows.

Holzhaus:
I think the needs some more discussion, I'll close this for now.

PR #240, opened 2021-05-08
Remove tools/update_avatars.py

Holzhaus:
This script is out of date an obsolete. Just building the website
locally should download the avatar to the correct directory
automatically.


daschuer (approved):
Thank you.

PR #243, opened 2021-05-09
pre-commit: Optimize SVGs automatically

Holzhaus:


tools/optimize_svg.py
```diff
@@ -0,0 +1,35 @@
...
+    args = parser.parse_args(argv)
+    options = scour.parse_args(list(ARGS))
+
+    for input_file in args.input_file:
+        with open(input_file, mode="rb") as input:
+            with tempfile.NamedTemporaryFile(
+                suffix=".svg", mode="wb", delete=False
+            ) as output:
+                scour.start(options, input, output)
+        shutil.move(output.name, input_file)
```
Swiftb0y: Can we process them in parallel?

Swiftb0y: Probably not really worth the trouble.

Holzhaus: I mean it takes a little while if you with `--all-files`, but usually only the files you commit are checked.

But I can look into it.

Swiftb0y: Thats probably already too much work. I guess it would also possibly introduce bugs if some entries are separate for some reason, etc. Disregard the first comment.

Swiftb0y (approved):

tools/optimize_svg.py
Holzhaus: Too late, check the latest commit :D

Swiftb0y (approved):

ronso0:
CI passed. merge?

PR #245, opened 2021-06-03
Update 2.3 screenshots

Holzhaus:


Swiftb0y:
Most of the image files don't seem to be used anywhere as part of this PR. Am I just not finding where they are used, or are you planning to use them later?

Holzhaus:
Used on the Screenshots and Features page.

Swiftb0y (approved):
Thanks, I only looked at the screenshot page, I wasn't aware there were images on the features page as well.

ronso0:
sorry I join in that late..
Is there a reason there are no samplers shown?
Though, good idea not to show the same view with all skins (minimal Tango decks, no effects row in Shade for example)

PR #251, opened 2021-08-11
remove $ and # prefixes from commands on download page

Be-ing:
These have to be deleted when copying and pasting commands into
a shell, which is inconvenient.

### DIFF
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -48,7 +48,7 @@ versions:
           The latest stable version of Mixxx is [available](https://archlinux.org/packages/community/x86_64/mixxx/) in the community repository and can be installed using `pacman`.
           Open a terminal, and enter:
 
-              pacman -S mixxx
+              sudo pacman -S mixxx
       - slug: source
         name: Source Code
         icon: terminal.svg
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -23,9 +23,9 @@ versions:
           We provide a [PPA on Launchpad](https://launchpad.net/~mixxx/+archive/mixxx) to make installing install the latest stable version of Mixxx as easy as possible.
           Open a terminal, and enter:
 
-              $ sudo add-apt-repository ppa:mixxx/mixxx
-              $ sudo apt update
-              $ sudo apt install mixxx
+              sudo add-apt-repository ppa:mixxx/mixxx
+              sudo apt update
+              sudo apt install mixxx
 
           Using the PPA ensures that new package versions will be installed automatically with `apt`. Otherwise, you can [download individual packages](https://launchpad.net/~mixxx/+archive/ubuntu/mixxx/+packages) and install them manually.
         note: |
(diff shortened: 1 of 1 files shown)


content/pages/download.md
```diff
@@ -48,7 +48,7 @@ versions:
           The latest stable version of Mixxx is [available](https://archlinux.org/packages/community/x86_64/mixxx/) in the community repository and can be installed using `pacman`.
           Open a terminal, and enter:
 
-              # pacman -S mixxx
+              pacman -S mixxx
```
ronso0: like the above commands, this requires `sudo`, doesn't it, so let's add it

ronso0 (changes requested):
LGTM, except the pacman command

content/pages/download.md
Be-ing: good catch, fixed

ronso0:
Thanks!

Holzhaus:
These prefixes denote whether its a user shell or root shell. I think they were useful, but ok.

Be-ing:
I ran into this inconvenience testing the Mixxx build instructions in a fresh Fedora 34 VM. I was testing the build instructions for Tenacity and figured I would double check the build instructions for Mixxx while I was working in the VM. So I copied and pasted the command for enabling the RPM Fusion Free repository but had to delete the `$` prefix. Now users can easily select the entire line to copy and paste verbatim into a shell.

As for distinguishing between user and root shells, we prefix the commands with `sudo` so that should be clear.

Holzhaus:
sudo is not pre-installed on Arch, but I guess most people will install it anyway.

Be-ing:
Nothing is preinstalled on Arch :laughing:

daschuer:
Can we adjust the command on arch that it works by default? What are the steps to install Mixxx without sudo?
Is su available.
Than we can use 
`su -c "packman -S mixxx"`
Does that work?

Holzhaus:
By default you just end up with a root shell. No user accounts, you have to add them via `useradd` yourself, and can optionally install and configure `sudo`. So the previous pacman command in a root shell was the command that works "by default". But I don't think it's much of an issue, I doubt that any Arch User does not add a non-root account or does not know what `sudo` does.

Swiftb0y:
> These prefixes denote whether its a user shell or root shell. I think they were useful, but ok.

I've been using linux for 5+ years and I have not heard of this convention... I think indicating the type of shell is better done by pre-fixing any command that requires root with `sudo` or mentioning the type of shell in the explanation.

Holzhaus:
> I've been using linux for 5+ years and I have not heard of this convention...

- `$` is for posix shells (normal user) 
- `#` is for posix shells (root)
- `%` is for C shell/zsh

The big advantage of using this convention as prefix is that you not only know what kind of shell you need to use, but also if the line contains a complete command or if it was linewrapped. 

Users that don't know that they need to remove the dollar sign in front of the line probably shouldn't copy-paste random commands from the internet anyway. But I don't really care, we can just keep it as is. 

> pre-fixing any command that requires root with sudo

As I said, `sudo` is not pre-installed on Arch, but I guess it's okay. People know what it means and are probably able to replace it with `su -c` or `doas` if necessary.

> mentioning the type of shell in the explanation.

To easy to overlook that IMHO. On the other hand "permission denied" is pretty clear anyway.

PR #261, opened 2021-11-10
POC for a Controller Hotlist page

daschuer:
This is a prove of concept of a Controller Hotlist generated from user interactions on Discourse. 

What do you think about it. The magic formula to create the heat value for a controller is missing. I need a read only API key to continue here. 

In addition we need an ignore list to ignore topics not specific for a certain model and a filter that can mark controllers already supported. 

My idea is this magic formula:
Consider all data younger than three years. 
heat = ("number actions by user per controller" camped to 0 .. 2) * 3 - last action age form user in years. 

What do you think?

### DIFF
--- a/content/pages/controller-hotlist.md
+++ b/content/pages/controller-hotlist.md
@@ -1,0 +1,9 @@
+---
+title: Controller Hotlist
+template: pages/controller-hotlist
+controllers:
+  replaced_by_discourse_hotlist_py:
+    name: Hercules RMX 2
+    forum_link: https://mixxx.discourse.group/t/custom-midi-mapping-hercules-rmx-2-0-1-3
+    heat: 78
+...
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -127,11 +127,6 @@ NAV_MENU = (
                 "Navigation bar link to Mixxx Press page",
             ),
             MenuItem(
-                "/controller-hotlist",
-                "Hot-List",
-                "Navigation bar link to Mixxx Controller Hotlist",
-            ),
-            MenuItem(
                 "/contact",
                 "Contact & Team",
                 "Navigation bar link to Mixxx contact page.",
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -89,6 +89,7 @@ PLUGIN_PATHS = [
 PLUGINS = [
     "author_metadata",
     "download_metadata",
+    "discourse_hotlist",
     "draft_override",
     "md_yaml",
 ]
@@ -124,6 +125,11 @@ NAV_MENU = (
                 "/press",
                 "Press",
                 "Navigation bar link to Mixxx Press page",
+            ),
+            MenuItem(
+                "/controller-hotlist",
+                "Hot-List",
+                "Navigation bar link to Mixxx Controller Hotlist",
             ),
             MenuItem(
                 "/contact",
--- a/plugins/discourse_hotlist/__init__.py
+++ b/plugins/discourse_hotlist/__init__.py
@@ -1,0 +1,3 @@
+from .discourse_hotlist import register
+
+__all__ = ["register"]
--- a/plugins/discourse_hotlist/discourse_hotlist.py
+++ b/plugins/discourse_hotlist/discourse_hotlist.py
@@ -1,0 +1,61 @@
+import json
+import logging
+import datetime
+import urllib.request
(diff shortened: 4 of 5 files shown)


daschuer:
Preview: 
https://deploy-preview-261--mixxx-website.netlify.app/controller-hotlist/

Swiftb0y:
At the moment, it just seems like you'd take all the threads from the controller mapping topic and rank them based on their age and number interactions. However, not all topics are concerning specific controllers. How do you plan to filter out those that are not concerning controllers. And you'd also need to filter out the hardware that already has a mapping.

daschuer:
Yes, I plan to add a manual ignore list for these topics. Later we may consider to decorate all relevant threads with a discourse tag or something. Since this is a static side we can afford some manual tasks for the first evaluation IMHO.

daschuer:
We need also consider controllers with two topics.

Swiftb0y:
I feel like the best option would be to just tag all threads that qualify for the hotlist. Then the python script would just need to query all topics that have this tag attached. Discourse even has a feature to tag topics automatically based on words in the post. The duplicate topic issue would also be solved by this allow-list approach.

daschuer:
Yes, that sounds reasonable.

Who with staff privileges can help here?

Holzhaus:
Let's first take a step back and talk about the purpose of such a page before we discuss details how to calculate the "heat", etc.

As a user, I'd be interested in 3 things:

1. What controller should I buy when I intend to use Mixxx? 
2. Is the controller I already own supported by Mixxx?
3. If it isn't supported, where can I get help for mapping it myself?

For 1/2., you can just check the manual or ask on the forums. For 3., we have the wiki, zulip and the forums.

I don't really get what purpose the "hotlist" serves, because it includes both supported controllers and unsupported controllers.

If it's just for us to know what controllers our users are interested in, I think the website is the wrong place. Also, we already exceed discourse page view limits, let's not add a bunch of additional crawler traffic everytime the website is built.

daschuer:
The use cases you mentioned are already covered by various links on the website. 

The main purpose of this list is to inform the core team about controller market trends. This can be used to derive activities like getting in touch with controller manufacturers and back this up with some trust-able data. 

The original discussed explicit voting system was, considered as not optimal because of the user expectations it may create.  From this perspective this can be a hidden page. On the other hand it would be fair that everyone can have a look at the data. 

I have picked our website as a home, because we have all facilities at a hand to create such a page. 

In the final solution this cannot be run on every website build. We need an extra trigger to build it on demand. For the normal builds this should use cached data from a controller-hotlist.md file. 

Not sure if the should be cron job or just a manual action.

Compared to crawlers that request all post via the html interface, we just need some metadata via the json API here, that shall be smaller by a magnitude. But if it turns out to be an issue anyway we can consider to introduce some caching that we only need to requests changes since the last run. 

I am really curious what would is the top rated controller according to the formula. 

Does this make sense?

Be-ing:
I share @Holzhaus's concern with putting this on the website. I think this would be misleading. I thought we had agreed that the content for the website or wiki would be a short, manually curated list of well supported controllers for different price ranges.

daschuer:
As I have already outlined it above: We can hide it. 
The main purpose is to inform the core team about controller trends.

Be-ing:
Okay then please remove this new page from the navigation menu.

Holzhaus:
> The main purpose is to inform the core team about controller trends. 

Then it should be either a bot that posts on Zulip or the forums on a regular basis (montly? Quarterly?) , or just a script in the `tools` directory that a developer can run locally and prints to the termibal.

The former could be implement using GH Actions, which supports a cronjob-like workflow AFAIK.

daschuer:
OK, I think we have got a common idea.  

For the sake of easy implementation I will prepare a stand alone Python script that calculates the hotlist. 
We can decide later If we just post the results on Zulip or give it a home somewhere else. 

Can one of the Discourse Admins create me a API key? I will limit the script to a small number of queries during debugging to not raise our hit count significant.  

What's next: 
* Lets also introduce three tags for the controller mappings category. Something like:   
  * Official Mapping 
  * Community Mapping
  * No Mapping
* Add or extend the sticky not to encourage users to create topic for controllers without a mapping to express the demand. 

What do you think?

daschuer:
I like to revive this task. Who can provide me an API key, that is a mandatory step, regardless to the question where to put the data.

PR #265, opened 2022-03-09
CI: Fix quoting of pelican --extra-settings argument

Holzhaus:


daschuer (approved):
Thank you.

daschuer:
Merging this has somehow triggered an deployment issue with Netifly. www.mixxx.org is now displayed without styles. 
Do you have an idea how to fix it? 

The issue is probably caused by the latest changes in pelican: 
https://docs.getpelican.com/en/latest/changelog.html

PR #273, opened 2022-06-13
fix(authors): fix previously broken default_avatar image

Swiftb0y:
Before: https://mixxx.org/news/author/matthew-nicholson/
After: https://deploy-preview-273--mixxx-website.netlify.app/news/author/matthew-nicholson/


ronso0:
so, what was the issue? avatar folder was moved?

Swiftb0y:
Yes. I moved the default_avatar.png back to the `theme/` directory because IMO its more theme-bound than actual user-supplied content.

Holzhaus (approved):

PR #279, opened 2022-09-03
content/news: add GSoC 2022 Work Product

davidchocholaty:
This PR adds the final blog with the work product for the Google Summer of Code 2022. The blog is written for the Pitch Shift effect and Group delay handling project which was implemented by davidchocholaty contributor.

PR #280, opened 2022-09-07
Use default Python version 

daschuer:
This is currently 3.7 on the current Xenial deploy server and will change to 3.8 after upgrading to the new Focal server.

The differences can be tracked in 
https://github.com/netlify/build-image/blob/focal/CHANGELOG.md


Holzhaus (changes requested):
It's good practice to pin dependencies. Otherwise stuff may  randomly break when the server switches versions but our code doesn't.

Could you please explain the rationale for this?

daschuer:
The Xenial deploy server will be switched off at 2022-11-15 

The switch to Focal fails due to this message:
```
8:45:21 AM: /opt/build-bin/run-build-functions.sh: line 223: /opt/buildhome/python3.7/bin/activate: No such file or directory
8:45:21 AM: Error setting python version from runtime.txt
```

https://app.netlify.com/sites/mixxx-website/deploys/63183dfa49bd2f000958bb88

The idea was to remove this version pinning to have our scripts running on Xenial and Focal to are able to switch forth and back during sorting out issues. 

Unfortunately we have no common version we can Pin 3.7 vs 3.8: 
https://github.com/netlify/build-image/blob/xenial/included_software.md
https://github.com/netlify/build-image/blob/focal/included_software.md

daschuer:
Now it fall back to 2.7. That's stupid.

daschuer:
Ok, so let's wait for https://github.com/mixxxdj/website/pull/279 and than pin 3.8 instead.

davidchocholaty:
> Ok, so let's wait for #279 and than pin 3.8 instead.

Thank you. It will be ready to merge probably in the upcoming 2 days.

Swiftb0y:
I guess this can be closed now?

daschuer:
Yes, superseded by #283

PR #287, opened 2023-01-05
add links to mastodon including verification

ywwg:


ywwg:
precommit failure looks unrelated to the change

theme/templates/base.html
```diff
@@ -25,6 +25,9 @@
     {# should be equivalent to rel=canonical -- used for aggregating likes #}
     <meta property="article:publisher" content="https://www.facebook.com/pages/Mixxx-DJ-Software/21723485212"/>
 
+    <!-- Mastodon verification backlink -->
+    <a rel="me" href="https://floss.social/@mixxx">Mastodon</a>
+
```
Swiftb0y: Does this really have to go in the header? It creates an ugly artifact. Wouldn't it make more sense to add this as a separate entry in the Community links at the bottom? Or Just add `display: none` so it gets hidden. 
Screenshot from 2023-01-05 22-35-12

Holzhaus:
Would be nice to add an icon + link to the header, next to the other social icons (Twitter, Facebook, Zulip, GitHub).

Holzhaus:
An icon can be found here: https://commons.m.wikimedia.org/wiki/File:Mastodon_Logotype_(Simple).svg

Probably just need to change the color to white using Inkscape.

theme/templates/base.html
ywwg: oh sorry, I thought it was a meta so it wouldn't appear.  I'll fix

Swiftb0y (approved):
Thanks LGTM. Objections @Holzhaus?

Holzhaus (approved):
LGTM, thank you!

PR #290, opened 2023-02-24
Add new post "Mixxx accepted for GSoC 2023!"

daschuer:

PR #292, opened 2023-02-28
Warn for non XXXX-XX-XX-my-post-title.md news articles

daschuer:
This adds a warning that is issued  during a PR if the file is not correctly named and the update_news_date.py will fail after merging to the website branch. 

This PR also aligns the used reg-expression, to enure the results are matching. 

I have prepared here a PR with a good and a bad example. 
https://github.com/daschuer/website/actions/runs/4297143440/jobs/7489749558

This should help to avoid noticing a deployment issue early before merge. 
This has happened here: https://github.com/mixxxdj/website/pull/290


Swiftb0y (approved):
LGTM. I think this is a helpful addition for troubleshooting issues likely occurring for guest authors.

PR #295, opened 2023-05-12
Download link for ARM macOS 

daschuer:
This should add the link to download ARM macOS. 
It also tries to detect the user agent architecture to download the right binary when pressing the download button.

### DIFF
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -14,6 +14,8 @@ versions:
             name: 64-Bit
       - slug: macos
         os: macOS 10.12 or later
+        text: |
+          Download Mixxx 2.3.5 for Intel macOS 10.12 or later. This also runs with Rosetta 2 on ARM macOS (Apple silicon). Alternatively use the the native ARM build development snapshot from below.  
         packages:
         - slug: macosintel
           name: Intel
@@ -85,6 +87,8 @@ versions:
             name: 64-Bit
       - slug: macos
         os: macOS 10.12 or later
+        text: |
+          Download Mixxx 2.5-alpha for Intel macOS 10.12 or later or for ARM macOS 11.0 or later (Apple silicon M1/M2). Check your OS by clicking the Apple Logo in the menu bar, then About this Mac -> If chip contains Apple M..., then download the ARM build.
         packages:
         - slug: macosintel
           name: Intel
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -88,7 +88,7 @@ versions:
       - slug: macos
         os: macOS 10.12 or later
         text: |
-          Download Mixxx 2.5-alpha for Intel macOS 10.12 or later or for ARM macOS 11.0 or later (Apple silicon M1/M2). Check your OS by clicking the Apple Logo in the menu bar, then About this Mac -> If chip contains Apple M..., then download the ARM build.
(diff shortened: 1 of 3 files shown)


Swiftb0y (commented):
LGTM, though I can't test the x86 vs ARM autodetect. Doesn't @fwcd have access to an ARM mac?

theme/static/js/download.js
```diff
@@ -23,9 +23,12 @@
         if (window.navigator.userAgent.indexOf("Windows") !== -1) {
             return ["windows", "win" + detectBitness()];
         }
-        if (window.navigator.userAgent.indexOf("Mac") !== -1) {
+        if (window.navigator.userAgent.indexOf("Intel Mac") !== -1) {
             return ["macos", ["macosintel"]];
         }
+        if (window.navigator.userAgent.indexOf("ARM Mac") !== -1) {
```
fwcd: Unfortunately, we can't use the user agent string for that, even native browsers identify as `Intel` on Apple Silicon Macs. For example, Firefox uses

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/113.0
```

and Safari

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Safari/605.1.15
```

This is presumably because too many websites have been using this to test against mobile browsers, so it's kept around along with all the other legacy stuff (sigh). Unfortunately, even `navigator.platform` reports as `MacIntel`.

The other options unfortunately aren't really satisfactory either:

- [Detect the WebGL renderer](https://stackoverflow.com/a/64958299): Chrome-only
- [Use similar WebGL hacks](https://stackoverflow.com/a/65412357): Doesn't work in Safari, for fingerprinting prevention reasons
- [Use `navigator.userAgentData`](https://stackoverflow.com/a/75177111): Probably the cleanest solution, but [not supported](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues) in Firefox and Safari yet

If everything fails, we'd have to present two download links. The [GIMP homepage](https://www.gimp.org/downloads/) seems to use similar tricks, which works in Firefox and Chrome and falls back to displaying buttons for both architectures in Safari.

fwcd: If we present two buttons we might want to consider adding a quick note on how the user can identify their architecture (e.g. clicking the Apple Logo in the menu bar, then About this Mac -> If chip contains Apple M..., then it's ARM)

daschuer:
Now it no longer immediately start the download on macOS. It scrolls to the details, where I have added the suggested information text.

content/pages/download.md
```diff
@@ -85,6 +87,8 @@ versions:
             name: 64-Bit
       - slug: macos
         os: macOS 10.12 or later
+        text: |
+          Download Mixxx 2.5-alpha for Intel macOS 10.12 or later or for ARM macOS 11.0 or later (Apple silicon M1/M2). Check your OS by clicking the Apple Logo in the menu bar, then About this Mac -> If chip contains Apple M..., then download the ARM build.
```
fwcd:
```suggestion
          Download Mixxx 2.5-alpha for Intel macOS 10.12 or later or for ARM macOS 11.0 or later (Apple silicon M1/M2). You can check which version you need by clicking the Apple logo in the menu bar, then "About this Mac". If the window displays an "Apple M..." chip, download the ARM build, otherwise you need the Intel build.
```

Wdyt?

daschuer:
Thank you.

Swiftb0y (approved):
Thank you. LGTM. Thank you very much @fwcd for sharing your expertise. I highly appreciate it.

PR #307, opened 2023-11-21
Update Arch Linux package URL in download.md

felixonmars:
The old URL returns 404 now.


daschuer (approved):
Thank you for keeping the head up. LGTM

PR #312, opened 2024-01-17
Add .build directory to .gitignore

uklotzde:
...and omit trailing slash from directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
-/output/
+# Directories without trailing slash to allow symbolic links
+/output
 *.pyc
 *~


Swiftb0y:
whats producing the `.build` folder for you? I don't have that when I build normally.

uklotzde:
> whats producing the `.build` folder for you? I don't have that when I build normally.

I opened the project in VS Code and the `/.build` directory was generated implicitly.

Swiftb0y:
cannot reproduce. do you have any python/pelican-specific vscode extensions installed (apart from the microsoft stock one)?

uklotzde:
After uninstalling this YAML extension by RedHat the `/.build` directory didn't appear again.

image

uklotzde:
Why is the `__pycache__` directory under version control?

Swiftb0y:
good question. I don't think it mattered yet because all the contained files where just `*.pyc` but its probably better to just put the __pycache__ into the gitignore instead. wydt?

PR #316, opened 2024-01-27
fix(theme): Fix video plugin always using video ID from first video

Holzhaus:
See: https://mixxx.zulipchat.com/#narrow/stream/248802-website/topic/2.2E4.20Release.20blogpost.20-.20Pelican.3A.20Multiple.20videos.20do.20not.20work/near/418451245

Note that you might need to clear your browser cache.


JoergAtGithub:
Works for me! Thank you!

PR #321, opened 2024-02-11
Update features page deck and skin example screenshots for 2.4

djantti:
Hi all! Here are updated 2.4 screenshots for the features page. All images should be the same size and style as the old ones. The LateNight skin example was changed to default PaleMoon variant and the top deck image has the new sync leader button activated.

Hopefully the images are good enough and this is the correct way to update them.


ywwg:
oh I'm sorry I didn't see this pull request before merging a similar one.  Can you resolve the conflicts? Your feature-deck image looks sharper than mine so let's go with that

djantti:
No worries! I'm still terrible with git, so here goes nothing... and apologies in advance. :sweat_smile:

PR #324, opened 2024-02-12
2.4 screenshots

ywwg:
pointing to main branch to get a preview


ywwg:
pointing back to 2.4

ronso0:
> pointing to main branch to get a preview

Ha, that's the trick!

LGTM, though the "50%" previews can be much smaller, for me they're rendered at ~300x180 px and I can watch them for a few seconds being downloaded.

Tbh, while it may make sense to show the window title bar (now playing text), the white top bars are distracting.
I don't want you to take the screenshots again, so can we simply cut that off?

ywwg:
yup I can crop those

ronso0 (approved):
Looks good, thank you.

Only issue (and we can't do much about it) is that Shade looks shitty when scaled up.
The other screenshots are excellent.
image

PR #328, opened 2024-02-16
move features link to features teaser section

ronso0:
currently it's in the Source Code container :shrug: 

image

### DIFF
--- a/theme/templates/pages/index.html
+++ b/theme/templates/pages/index.html
@@ -84,24 +84,27 @@
       </p>
     </div>
   </div>
+
+  <div>
+    <p>
+      {% trans %}See dozens <a href="{{ SITEURL }}/features">more features</a> or download Mixxx and start DJing now!{% endtrans %}
+    </p>
+  </div>
 </div>
 
 <div id="opensource" class="docs-section cover-bg">
   <div class="container splash-text">
     <h2><a class="headline-link" href="#opensource">{% trans %}Free &amp; Open Source{% endtrans %}</a></h2>
     <p>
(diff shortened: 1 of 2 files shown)


Swiftb0y:
you'll probably also want to change the enclosing div container to be `thirds` (or something idk, not familiar with the css framework) instead of `quarters` so the buttons are still centered. 
image

ronso0:
yes, I'm still tweaking it, have no clue about the css rules here either.

theme/templates/pages/index.html
```diff
@@ -83,14 +83,20 @@ <h5 class="docs-header">{% trans %}Vinyl Record Control{% endtrans %}</h5>
         {% trans %}Use turntables with timecode vinyl records to control playback and scratch your digital music files as if they were pressed on vinyl. Mix music vinyls into your set by toggling vinyl passthrough mode.{% endtrans %}
       </p>
     </div>
+
+    <div>
+      <p>
+        {& trans %}See dozens more features or download Mixxx and start DJing now!{% endtrans %}
```
Swiftb0y: typo ;)
```suggestion
        {% trans %}See dozens more features or download Mixxx and start DJing now!{% endtrans %}
```

ronso0:
why is the Download button blue in your screen?
should be orange
image

Swiftb0y:
> why is the Download button blue in your screen?

That has always been like that for me for deploy-previews. Isn't that a feature? It's fine when I build locally and fine on prod.

ronso0:
> you'll probably also want to change the enclosing div container to be thirds

Hmm, will test. maybe the spcing is too much and the buttons look lost.

Swiftb0y:
> That has always been like that for me for deploy-previews. Isn't that a feature? It's fine when I build locally and fine on prod.

Ah, no. I see. It's dark-reader which is disabled for localhost as well as plain mixxx.org, but it does interfere in the deploy previews.

ronso0:
> > you'll probably also want to change the enclosing div container to be thirds
> 
> Hmm, will test. maybe the spcing is too much and the buttons look lost.

The buttons are great in the Source section but I think here an inline link suits better.

ywwg (approved):

PR #345, opened 2024-04-24
(fix) add #translators anchor, relevant for mobile view only

ronso0:


daschuer (commented):
I can confirm this  issue and that it is fixed now. Thank you.

PR #346, opened 2024-05-02
Fix typos in DVS internals (pt. 2) post

fwcd:
Some typos I found in the DVS pt. 2 blog post

### DIFF
--- a/content/news/2021-12-22-dvs-internals-pt2.md
+++ b/content/news/2021-12-22-dvs-internals-pt2.md
@@ -239,11 +239,11 @@ The Serato Control CD has a play time of approximately 16 minutes 20 seconds, or
 At a timecode frequency of 1000 Hz (cycles per second), we have 980,000 cycles.
 Every cycle encodes one bit, therefore we need an LFSR with an output length of at least 980,000 bits before it starts to repeat.
 
-At most, an n-bit LFSR can output 2<sup>n - 1</sup> bits before it starts to repeat.
+At most, an n-bit LFSR can output 2<sup>n</sup> - 1 bits before it starts to repeat.
 An LFSR with such a period size is called a maximal-length LFSR, and we already mentioned that Serato states that it uses a such an LFSR on the packaging of its timecode media.
 
 The smallest possible LFSR that can output at least 980,000 bits without starting to repeat needs at least 20 bits of state.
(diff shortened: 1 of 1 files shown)


fwcd:
I think there's some typos in the math too (under "How big does the LFSR need to be"). Would appreciate a second pair of eyes checking that though. The article states that a maximum-length n-bit LFSR has a period size of $2^{n - 1}$:

> At most, an n-bit LFSR can output 2<sup>n - 1</sup> bits before it starts to repeat.

According to Wikipedia this should be $2^n - 1 (\neq 2^{n - 1})$ instead.

> A 19-bit LFSR is too small, because 2<sup>19</sup> - 1 = 524,288 is less than 980,000...

The calculation has a typo too: $2^{19} - 1 = 524,287 \neq 524,288$.

content/news/2021-12-22-dvs-internals-pt2.md
```diff
@@ -239,11 +239,11 @@ The Serato Control CD has a play time of approximately 16 minutes 20 seconds, or
 At a timecode frequency of 1000 Hz (cycles per second), we have 980,000 cycles.
 Every cycle encodes one bit, therefore we need an LFSR with an output length of at least 980,000 bits before it starts to repeat.
 
-At most, an n-bit LFSR can output 2<sup>n - 1</sup> bits before it starts to repeat.
+At most, an n-bit LFSR can output 2<sup>n</sup> - 1 bits before it starts to repeat.
```
Swiftb0y, line 242: lmao, wikipedia is also wrong on this. https://en.wikipedia.org/wiki/Linear-feedback_shift_register#Output-stream_properties

Swiftb0y (approved):
LGTM. makes sense. thanks for double checking everything

content/news/2021-12-22-dvs-internals-pt2.md
fwcd: Interesting, I think it's correct in other sections of the article. Not that this would make it any less confusing... I usually don't feel confident enough to edit math-related things on Wikipedia myself, but here it might actually make sense to correct it 😄

Swiftb0y:
fyi just looking for the lgtm from you @Holzhaus since its your post.

Swiftb0y:
Jan seems busy. :shrug:

PR #350, opened 2024-05-09
2.4.1 release update

ronso0:
### TODO
- [x] create https://downloads.mixxx.org/releases/2.4.1/manifest.json
- [ ] trigger RPM Fusion builds?

### DIFF
--- a/content/pages/download.md
+++ b/content/pages/download.md
@@ -3,9 +3,9 @@ title: Download
 template: pages/download
 versions:
   stable:
-    name: 2.4.0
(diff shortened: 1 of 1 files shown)


daschuer:
The manifest should have been created automatically.
Unfortunately something with checkout failed and it created a 2.4.0 manifest.
I have restarted the build. Hopefully it works now. 
It is bad, that all was green in that case. Unfortunately I have no idea how to fail in that case. Let's investigate.

daschuer:
Works now :-)

PR #363, opened 2024-09-21
Add release roadmap and news to introduce the page

acolombier:
Note that the CSS still needs to be cleaned up from some of the weird class name.

- [Page](https://deploy-preview-363--mixxx-website.netlify.app/release_roadmap/)
- [News](https://deploy-preview-363--mixxx-website.netlify.app/news/2024-09-22-previsional-release-roadmap/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /output
 # python bytecode cache
 __pycache__
+.venv/
--- a/theme/templates/pages/release_roadmap.html
+++ b/theme/templates/pages/release_roadmap.html
@@ -1,0 +1,149 @@
+{% extends "base.html" %}
+
+{% block title %}{% trans %}Previsional Release Roadmap{% endtrans %}{% endblock %}
+
+{% block extra_head %}
+<link rel="stylesheet" href="{{ SITEURL }}/theme/css/pages/release_roadmap.css">
+
+<script src="https://cdn.jsdelivr.net/npm/mermaid@11.2.1/dist/mermaid.min.js" integrity="sha384-7J/UM6Vde6cXqst3EI86P36213Jz5rooW1EEUvweQ3F3gQOHkTuSKAX67ANmYljn" crossorigin="anonymous"></script>
+
+<script type="text/javascript" nonce="VoYlObL+iS4p0ATVpB4f495PwA9FE4obmpj2VWc961Q=">
+  mermaid.initialize({ startOnLoad: true });
+</script>
+
+{% endblock %}
+
+
+{% block body %}
+<div class="docs-section container">
+  <h1>Previsional Release Roadmap</h1>
+
+  <p>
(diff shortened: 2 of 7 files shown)


JoergAtGithub:
This looks nice, but I think we shouldn't publish future release dates, as our development model is not calendar based.
Can we add a text there like "60days after 2.5.0" or "minimum 90 days" instead of the artificial absolute dates?

.gitignore
```diff
@@ -2,3 +2,4 @@
 /output
 # python bytecode cache
 __pycache__
+.venv/
```
Swiftb0y, line 5: I think this rather belongs into your personal gitignore. There is no established convention on how call the venv directory (I just call it plain `venv` and have that in my gitignore).

theme/templates/pages/release_roadmap.html
```diff
@@ -0,0 +1,149 @@
+{% extends "base.html" %}
+
+{% block title %}{% trans %}Previsional Release Roadmap{% endtrans %}{% endblock %}
+
+{% block extra_head %}
+<link rel="stylesheet" href="{{ SITEURL }}/theme/css/pages/release_roadmap.css">
+
+<script src="https://cdn.jsdelivr.net/npm/mermaid@11.2.1/dist/mermaid.min.js" integrity="sha384-7J/UM6Vde6cXqst3EI86P36213Jz5rooW1EEUvweQ3F3gQOHkTuSKAX67ANmYljn" crossorigin="anonymous"></script>
```
Swiftb0y, line 8: not a fan of having rendering the static content client side all the time. I'd be nice if there was a pelican plugin that rendered this inline at build time... Wrapping https://github.com/ouhammmourachid/mermaid-py may be feasible. We already have other plugins (such as the privacy-friendly youtube embed) that could be used as an example.

acolombier:
> This looks nice, but I think we shouldn't publish future release dates, as our development model is not calendar based. 

The reason for me putting this document together is precisely because I am lost with these relative dates. My initial aim was to build a document that would help me to manage my time and efforts on various feature. This doesn't mean I want to create a surge of feature to review just before the release, but rather organise which one should be addresses together as they are closely linked (e.g STEM PRs and S4 Mk3 screens should ideally go together). We already have the milestone to group them, but there is no indication on calendar for that matter.

Note this is why I have added that large warning box to inform that these dates are only here as an indication. I guess to make it clearer, I could remove the day in it as this clearly is irrelevant (just `year/month`)

theme/templates/pages/release_roadmap.html
acolombier: Unfortunatelly, this Python plugin doesn't support Timeline and Gitgraph. However, I agree it is definitely sub-optimal to address that on client side. One solution could be to use the (Node-based) mermaid-cli which allows you to generate SVG out of the diagram, so this could be added in a `Makefile` for example and/or the CI. WDYT?

Swiftb0y: yeah neither are great. I'd almost prefer the current client side solution then I guess.

PR #370, opened 2024-11-27
Fix 2.4.2 link in download page

daschuer:
We have an issue that the release blogpost is not deployed for some reason. 

Netlify states: 

> All files already uploaded
> All files already uploaded by a previous deploy with the same commits.

Keep investigating


daschuer:
Preview works. Hopfully the deployment works as well.

PR #372, opened 2024-11-27
fix: replace `\r\n` in blogposts with `\n` after merge

### LINKED ISSUE #371: update_new_date.py bug with /r/n files
There was a major hassle with the release block post, because our update_new_date.py was not able to handle /r/n formated files.

I think it is "normal" that blog writers just open a file with their favorite editor without taking care of encoding. Changing the encoding is not an easy task depending on the used editor. 

How do we want to fix it? Shall we add /r/n support to the script? Or shall we convert the Ile in the fly?

Swiftb0y:
This is done to consistently enforce the Unix `\n` (linefeed) line ending convention in the repo even if blogposts where merged with Windows `\r\n` line endings. It also allows the script to easily appropriately handle such files, as before it was unable to properly parse it because it assumed files always were `\n`.

Fixes #371

### DIFF
--- a/tools/update_news_date.py
+++ b/tools/update_news_date.py
@@ -133,7 +133,10 @@ def main(argv=None):
                     continue
 
                 with open(
-                    filepath, mode="r", encoding="utf-8", newline="\n"
+                    filepath,
+                    mode="r",
+                    encoding="utf-8",
+                    newline=None,  # translate any \r\n to \n during read
                 ) as fp:
                     header, sep, body = fp.read().partition("\n\n")
                     if not RE_POST_STATUS.findall(header):


.github/workflows/website.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - uses: EndBug/add-and-commit@v9
       if: env.GIT_DIRTY != null && env.GIT_DIRTY != '0'
       with:
-        add: 'content/news'
+        add: '-A content/news'
```
daschuer: Is this change a debug leftover?

Swiftb0y: ~~No its required because in the case where the script does the line ending fixing, it essentially changes the entire file. With the renaming in the same step, git will think that the `XXXX-XX-XX-*` file has been deleted and a new one created. The `-A` ensures that it stages the "new" file as well as the "deleted" one. At that point it also realizes again that that was in fact just a renaming:~~

I just tested it again (this time passing `content/news` instead of `content/news/*`) and it picked it up. So this is in fact unnecessary. I will remove.

daschuer (approved):
Thank you.

Swiftb0y:
Thank you.

PR #374, opened 2024-12-17
feat: add static Christmas theme

acolombier:
Adding a static Christmas theme (logo+snow fall), align with the initiative in mixxxdj/mixxx#14027

This should be revert after the 6th of Jan.

Santa hat license to be approved.


Holzhaus:
Can you rebase on latest `website` branch? I think that should probably fix the build errors.

acolombier:
Ah I had just fixed it :sweat_smile:

acolombier:
@Holzhaus I have pushed my fix as it implied to use lastest version rather than reverting to a [fairly old version of PyYAML](https://pypi.org/project/PyYAML/5.3.1/) - what your thought about this? Would you like to resync your fork?

Edit: looks like there is more work to support newer version of PyYAML, but we should really look into it as the current version is [quite vulnerable](https://www.cvedetails.com/vulnerability-list/vendor_id-13115/Pyyaml.html)

Holzhaus:
I agree we should try to use a newer version, but I'd say the risk is rather low since we do not parse zntrust input, only stuff from this repo. And the Netlify build container is also isolated.

Maybe we can switch to `ruamel.yaml`? IIRC it also faster.

Holzhaus:
Re: This PR: the Mixxx logo looks slightly transformed. Aspect ratio seems to have changed, it's less wide but has roughly the same height compared to the logo without this change. But doubt someone notices without direct comparison ;)

Looks cool, thanks!

acolombier:
My fear is for someone to submit a PR to this repo, which would triggers the CI and potentially steal token to perform privilege escalation, similar to [Ultralytics](https://github.com/ultralytics/ultralytics) event [last week](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection), The risk is very low, and I believe first time contributor need CI approval so we should be fine. Still worth not letting out for too long :) 

> Maybe we can switch to ruamel.yaml? IIRC it also faster.

I've never used this one before, but no objection! Anything maintained would be good to me.

> Looks cool, thanks!

Glad you like it! Just need the :+1: on the hat used due to potential license risks

daschuer:
> Just need the 👍 on the hat used due to potential license risks

Mmm that could be really an issue,  can't we just draw our own?

Eve00000:
Very nice. I appreciate the work done.
It's fun and creates a christmas-community-spirit.

acolombier:
>  can't we just draw our own? 

I'll see what I can do!

acolombier:
Updated with @peineduper's asset rework as well! 

image

We can just press merge as soon as we want to turn on the XMas Mixxx mood

Eve00000:
> We can just press merge as soon as we want to turn on the XMas Mixxx mood

will there be some snowflakes behind the blogpost as well?

acolombier:
Yes, you can see a live preview [here](https://deploy-preview-374--mixxx-website.netlify.app/)

Eve00000:
> Yes, you can see a live preview [here](https://deploy-preview-374--mixxx-website.netlify.app/)

ah ok, sorry, I thought it would only be in the upper part as in the preview, because when I click on 'news' there are no snowflakes.
I like this 'temporary theme' a lot. ❤️ 🎅

acolombier:
I'm glad yo hear you do like it!
Oh sorry, I misread your comment about snowflakes. I wouldn't mind adding the snowflake falls on the news page as well! Just keen to hear what the rest of the team think about it?

theme/static/css/pages/index.css
```diff
@@ -30,3 +30,125 @@
 #opensource.cover-bg {
     background-image: url('/theme/images/cover/code-bg.jpg');
 }
+
+/* Temporary Christmas theme - thanks to https://pajasevi.github.io/CSSnowflakes/ */
```
Holzhaus: I recommend moving this to a dedicated file. Otherwise Browsers might decide the cached CSS is still valid and not show this ;-)

acolombier:
Merge?

Holzhaus (approved):
Feel free to hit merge yourself when you think the time fits.

daschuer:
I am already in Christmas mood, with all our efforts. Thank you.🎅🎄🌠

Eve00000:
❄️ ❄️ ❄️ 🎁 ☃️ I ❄️ am ❄️ getting ❄️ in ❄️ the ❄️ mood ❄️ as ❄️ well 🎅🎄  🎁 ☃️ ❄️ ❄️ ❄️ ☃️ 🎁 
thank you for creating this

acolombier:
Thanks for starting the Christmas train @daschuer ! Do you want to merge the PR now? I don't seem to have the permission to do so.

daschuer:
@acolombier I have just granted your the access to the website and manual

fwcd:
Love this, looks awesome!

PR #378, opened 2025-01-01
Reference the tagged 2.5.0 version for windows

daschuer:
It is now available after fixing signing issues


acolombier:
Tested and LGTM! Haven't tested the MSI as I don't have a Windows box available.
Thank you!

PR #392, opened 2025-05-16
Fix the mobile menu and prevent the page scrolling when the mobile menu open

### LINKED ISSUE #262: mobile menu can't be scrolled, Download not visible
instead I see the page itself is scrolled
image

Deepak-cell311:
screen

Closes #262

### DIFF
--- a/theme/static/css/custom.css
+++ b/theme/static/css/custom.css
@@ -23,7 +23,6 @@ html[data-theme="dark"] {
     --button-primary-color: white;
 }
 
-
 /* add a transition to ease into new themes */
 /* looks best in Mozilla, text transitions late in Safari/Chrome */
 html.transition-theme,
@@ -406,7 +405,6 @@ img.responsive {
     border-bottom: 1px solid var(--border-color-softer);
     padding: 2rem;
     max-height: unset;
-
 }
 
 @media only screen and (min-width: 600px) {
@@ -428,8 +426,6 @@ img.responsive {
         width: 100%;
         border: 0;
         padding: 0;
-    /* height: 100vh; */
-
     }
 
     .navbar-download-button .button {
@@ -755,10 +751,9 @@ article.news + article.news {
     margin: 0 0 2em 0;
 }
 
-
-
-.scroll-lock {
-    overflow: auto;
-    position: fixed;
-    width: 100%;
+#page-wrapper.scroll-lock {
+    overflow: hidden;
+    height: 100vh;
+    touch-action: none;
+  -webkit-overflow-scrolling: none;
   }
--- a/theme/static/css/custom.css
+++ b/theme/static/css/custom.css
@@ -406,7 +406,7 @@ img.responsive {
     border-bottom: 1px solid var(--border-color-softer);
     padding: 2rem;
     max-height: unset;
-    
+
 }
 
 @media only screen and (min-width: 600px) {
@@ -762,4 +762,3 @@ article.news + article.news {
     position: fixed;
     width: 100%;
   }
-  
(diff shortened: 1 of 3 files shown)


Deepak-cell311:
Hi @Swiftb0y 
Kindly have a look over and review this PR whenever you have a moment.
Thank you🚀

theme/static/css/custom.css
```diff
@@ -23,6 +23,7 @@ html[data-theme="dark"] {
     --button-primary-color: white;
 }
 
+
```
acolombier: nit: unnecessary change
```suggestion
```

theme/static/css/custom.css
```diff
@@ -404,6 +406,7 @@ img.responsive {
     border-bottom: 1px solid var(--border-color-softer);
     padding: 2rem;
     max-height: unset;
+
```
acolombier: nit: unnecessary change
```suggestion
```

theme/static/css/custom.css
```diff
@@ -425,6 +428,8 @@ img.responsive {
         width: 100%;
         border: 0;
         padding: 0;
+    /* height: 100vh; */
+
```
acolombier: nit: comment cruft
```suggestion
```

theme/static/css/custom.css
```diff
@@ -749,3 +754,11 @@ article.news + article.news {
     padding: 1em;
     margin: 0 0 2em 0;
 }
+
+
+
```
acolombier: nit: unnecessary spaces

```suggestion

```

theme/templates/base.html
```diff
@@ -65,5 +65,17 @@
...
+    <script>
+      const navToggle = document.getElementById('navbar-hamburger-button');
+
+      navToggle.addEventListener('change', () => {
+        if (navToggle.checked) {
+          document.body.classList.add('scroll-lock');
+        } else {
+          document.body.classList.remove('scroll-lock');
+        }
+      });
```
acolombier: `#navbar-hamburger-button` is [a hidden element](https://github.com/mixxxdj/website/blob/9521887c36e3292a040825f5bceaa01015aebe5e/theme/static/css/custom.css#L384) used to persist the expected state of the menu AFAIU, so that `scroll-lock` won't have any effect.

acolombier (commented):
Hi @Deepak-cell311, thanks for your contribution!

Could you clarify what does your change? I can see it has extended to menu drawer to fill the window, is there something else this is intending to fix?

(Before/after video)

https://github.com/user-attachments/assets/cf256421-67eb-4e80-b78a-ea41381ef1d3

Deepak-cell311:
Hi @acolombier,

Thank you for the review and helpful feedback.

In my previous implementation, I used the hidden checkbox input (#navbar-hamburger-button) to detect when the mobile menu was toggled and then added a scroll-lock class on the <body> to prevent the background from scrolling.

However, as you rightly pointed out, since the input is visually hidden and used only to preserve state, it doesn’t reliably control page scrolling across all browsers and screen readers. As a result, the page behind the open menu remained scrollable in some environments, which wasn’t the expected behavior. It only works on my local machine.

In my current approach, I've updated the structure to wrap the entire page content inside a #page-wrapper element. When the mobile menu is opened, the scroll-lock class is now applied to #page-wrapper instead of body. This gives more precise control over what gets locked and ensures the background remains fixed while the menu is open. I've also updated the CSS accordingly to disable overflow and enforce fixed height:

```
#page-wrapper.scroll-lock {
  overflow: hidden;
  height: 100vh;
  touch-action: none;
  -webkit-overflow-scrolling: none;
}
```
This approach keeps the scroll-lock behavior isolated, avoids unintended side effects, and works consistently across devices. Let me know if you’d like me to further tweak or simplify anything.

Thanks again!

PR #397, opened 2025-07-03
Internationalisation support to the website

JosepMaJAZ:
Work in progress to add language support to the website.

Even though it is true that nowadays some browsers give the option to autotranslate a site, giving a default translated site could better suit new and existing users.

This PR adds i18n_subsites pelican plugin which generates subfolders with translated content, and a dropdown list of supported languages has been added to the navigation bar.

Ideally the site should switch to the language that the browser reports, but so far I'm leaving that out of the equation.

Next step is checking the pages that need to add gettext/trans annotations. 

I'm not 100% sure if the markdown pages will be as easily translatable as the templates.
Also I need to envision a way to translate the menu/navigation bar, since currently it is defined in the pelicanconf.py, not in the html templates.

- [x] Generate subsites with i18n-subsites plugin
- [x] Incorporate a language selector (to the footer) 
- [x] Incorporate old translations into the code
- [x] Translate menu (since it's python code, maybe it could go directly into the template and see if then I can add gettext to translate it and gets exported to the .pot file)
- [ ] Add %trans% indicators to missing texts in templates.
- [ ] Verify that all text that is expected to be translated, gets translated (transifex makes this easy, in the english language, it generates a pseudo translation in "l33t c0d3"). Currently there are some places that are expected to be translated that don't.
- [ ] Update transifex to have the translations for the new site. This will require someone from the core team. The instructions should be very similar to how they are uploaded and downloaded for Mixxx.
- [ ] Once all is done, squash and move it to a branch, close this PR and create a new one so that the merge can be done properly.


acolombier:
This looks like a great adding! How are we planing to handle the blog post? Auto translation like GTranslate?

JosepMaJAZ:
We first need to verify if translating the posts (which are in markdown) will be possible.

But it might be a bit complicated, indeed, if each post needs to be translated, but at the same time, it could be easier for a reader to get new features in his language.

Automatic translations could be a start, so that translators don't need to start from zero, and just ammend incorrect translations

daschuer:
I don't think it is feasible to make the article translatable via transifex. We may integrate the web browser translation feature 
like:  
for other languages klick here:
image

for this we need to set the `<html lang="en">` correct. It is currently empty.

README.md
```diff
@@ -91,6 +115,31 @@ use in multiple places on the site, place it in the `/theme/templates/` folder
 All strings wrapped with `{% trans %}Hello World {% endtrans %}` are flagged
 for translation. Whenever adding new English strings to the website, please
 wrap them in a `{% trans %}Hello World{% endtrans %}` block.
+A more compact way is using gettext
+
+{% trans %}translatable content{% endtrans %}
+{{ gettext('translatable content') }}
```
daschuer, line 120: Both is printed in one line. I think you need to add a blank line. It is probably better to recommend only one translation method. I like `{ tr('translatable content') }` to match the mixxx source. But it doesn't really matter.

daschuer (commented):
Wow, Thank you very much for picking this up.

JosepMaJAZ:
Last commit has a somewhat working solution, but still misses the menu translation and the languages list is cropped if it can't be shown completely in the screen size.
Maybe we can move that to another place (at the footer, as a small list of languages?)

Also I realised i didn't create a new branch before creating the PR. Wonder if I should start a new one or it's ok now.

Eve00000:
Nice!
Would it be possible to only display languages that actually have a translation?
I can imagine it will be frustrating to select a language and than see everything in English as well.

JosepMaJAZ:
I am currently refitting the older translations from transifex into this and will put only those with a percentatge at or above 25% translated.

JosepMaJAZ:
Ok, now this is in a workable state.
Missing: 
- Translate the menu in some way (Currently the menu is defined in pelicanconf.py, not in the .html templates, so it is not extracted as translatable text). 
- Go check all templates to ensure that everything that needs to be translated has  the trans or gettext keywords.
- For some reason, the download page does not build correctly on windows. I've been skipping this for now, but I'll check it eventually.

JosepMaJAZ:
ok, the download page error seems to be an actual bug. %s is not a valid format in python:

                      File "C:\programacio\mixxxwebsite\plugins\download_metadata\download_metadata.py", line 54, in page_generator_context
                        manifest_url, {"timestamp": datetime.datetime.now().strftime("%s")}
                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
                    ValueError: Invalid format string

I've changed it locally to %S instead, but that should be solved in a separate PR . Seems this was done in 2021 by @Holzhaus

README.md
JosepMaJAZ: I fixed the line break, but I am unsure about keeping just one.
Also note that I cannot randomly choose a name, although I could define a function in the pelicanconf.py file that calls to the real one. 
Using "tr" would be confusing (the only thing in common would be that it is a translatable text)

Anyway, all the existing code uses {% trans %} {% endtrans %}, (except the site language text that I've added), but using the gettext syntax is problematic if one needs to escape text.
So, I am still in doubt of what we do. Maybe recommend the longer syntax, but that the shorter one can be used too?

Eve00000:
Nice work. 
I see translators will have a lot of work to get the site multilangual.
Maybe 25% translated is to low to show a language in the selector (not counting md)? Most languages don't have much more than translated titles (someone who started with good intentions).

Am I looking over a language indicator (other then in the url)?
Where did the 'Fake english' come frome?
The links to the manual in the different languages contains an error (except English).

JosepMaJAZ:
Hello @Eve00000 .

You're right that there are hundreds of lines to translate. This includes explanations of the features, or where and how to get in contact.

The older site had already translations, which is why I started the zulip thread and ended in this PR. Like Mixxx itself, it's in transifex, here: https://app.transifex.com/mixxx-dj-software/mixxxdj-website/dashboard/

I took that as a reference ( %) of the amount of translations done (although some texts no longer exist and there are some other new). Then, I "upgraded" the translations with the messages generated for the current site.

I need to readd the translations of the menu entries, since I removed them in the upgrade process because they weren't available yet.

Also I will need to remove the fuzzy indicator on the translations (basically means the translation is not revised, sometimes means that is autogenerated or copied from an already existing translation). If I don't do that, those are not considered as translated.

I am not sure of what you mean with language indicator. If you mean to know in the site which language is selected, the commit you probably tested shows the selected language in white instead of yellow.

The fake english is partially autogenerated from transifex. 
If you go into the english (reference) language, you can also export it, and then it generates this pseudolanguage in order to verify that translations work, and also adds some extra text that is intended to verify that there is enough space to show text longer than the original (you probably know that english is generally shorter than other languages). Then I manually "translated" the new lines.

Manual link: Oh, you're right. I forgot that this relative path is not part of the site, but another subsite. Fixed and will commit the change.

Eve00000:
> ... hundreds of lines to translate ... 

When you 're ready with you big summer clean-up and you have a po with all lines that can be translated, I think we can do a big effort to translate them in the major languages (I prefer to do it directly in the po as it speeds up a lot).

> I am not sure of what you mean with language indicator. If you mean to know in the site which language is selected, the commit you probably tested shows the selected language in white instead of yellow.

In the current version (deploy->preview) I don't see an indicator anymore, In an earlier version there was a language dropdown but that's gone, now the languages are all at the bottom (see images) so no white or yellow.

top
afbeelding

bottom
afbeelding

> The fake english is partially autogenerated from transifex ...  

Thank you for explaining, I was just wondering why it was integrated but it makes sense showing it for testing resaons in the PR

PS: still enjoying your free time? 😄

PR #398, opened 2025-07-25
Add a Traktor MK2 blog post

jclsn:
This blog post explains the current state of implementation for the Traktor MK2 decoder.

[preview](https://deploy-preview-398--mixxx-website.netlify.app/news/2025-08-27-dvs-internals-pt3/)

PR #399, opened 2025-07-25
Feat/render math plugin

Swiftb0y:
Since this will be useful for #398

[Preview](https://deploy-preview-399--mixxx-website.netlify.app/news/2025-07-25-render-math-test/)

### DIFF
--- a/content/news/XXXX-XX-XX-render-math-test.md
+++ b/content/news/XXXX-XX-XX-render-math-test.md
@@ -1,0 +1,11 @@
+title: Now featuring MathJax
+authors: Nikolaus Einhauser
+status: draft
+comments: yes
+math: yes
+
+Here's a pretty formula I like:
+
+$$ \sum_{i=1}^{n}{i} = {n(n + 1) \over 2 }$$
+
+also with inline $a_i$ variables
--- a/netlify.toml
+++ b/netlify.toml
@@ -48,5 +48,5 @@
 [[headers]]
  for = "/news/*"
  [headers.values]
-   # Additionally allow YouTube/Discourse frames and scripts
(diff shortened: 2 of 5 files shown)


Swiftb0y:
Interesting, this renders locally but not deployed. Likely a CSP problem. I'll look into it.

Swiftb0y:
So actually deploying the plugin (or rather mathjax loading from an external CDN) forces me to do an amount of CSP hole punching I'm not comfortable with.

jclsn:
So what do we do? I think math is an essential feature. It makes the post much nicer and more readable.

Eve00000:
on https://viereck.ch/latex-to-svg/
you can convert your formula to a svg

If you set the color to #FFFFFF and download it you get something like this:
equation

(this is an svg uploaded in my comment)

Swiftb0y:
I also favored the manual SVG approach until I remembered that inline math typesetting is also quite desirable to have (and not as easy to do with images). I wish there was a server side solution for this (as it would be the most elegant in this case), but there isn't afaict.

Swiftb0y:
@jclsn this works now (apart from rendering equations in the summary text, but thats a known upstream issue I don't want to look into).

content/news/XXXX-XX-XX-render-math-test.md
```diff
@@ -0,0 +1,11 @@
+title: Now featuring MathJax
+authors: Nikolaus Einhauser
+status: draft
+comments: yes
+math: yes
```
Swiftb0y, line 5: this is important, any article that wants to use math needs to opt into it (because I don't want to load it on every single article that doesn't need it)

Swiftb0y:
Merged through #398

PR #401, opened 2025-07-28
WIP: Final Report, GSOC 2025

ar-in0:
This PR introduces a blog-post describing the implementation of additional resamplers for Mixxx - a deliverable for the GSOC 2025 final evaluation.

[Preview](https://deploy-preview-401--mixxx-website.netlify.app/news/2025-07-30-gsoc-2025-report-armaan-chowfin)

PR #404, opened 2025-08-04
Final touches to the gsoc2025 blog post on resampling options for Mixxx

ar-in0:


ar-in0:
@daschuer 
With reference to your comments on the previous PR:
https://github.com/mixxxdj/website/pull/401#discussion_r2250435791

The relevant passage in the current version is:

---

#### The Mixxx Audio-Playback Stack
Mixxx exposes two important parameters in the *Sound Hardware Preferences* panel:

- **Sample Rate (Hz)**: Determines the DAC sample rate, i.e., how frequently audio frames are converted to analog signals.
- **Audio Buffer (ms)**: Specifies the total buffer duration, indirectly determining the size of ALSA’s ring buffer.

*Mixxx Sound Hardware Preferences*
Mixxx Sound Hardware Preferences

These parameters together influence the quality of output sound by defining the size in frames of the ALSA ring-buffer for the selected sample-rate:
```Ring Buffer Size = (Audio Buffer in s) * (Sample Rate in Hz)```

##### Mixxx Buffering Hierarchy
As audio frames move from Mixxx to the speakers, they pass through three distinct buffering levels:

**Mixxx Buffer** (userspace, heap):  
   Holds outbound frames from the track, possibly post-processing or effects.  
   This buffer is typically larger and acts as the staging area for frames handed off to ALSA.

**ALSA Ring Buffer** (kernel-managed, DMA-mapped):  
   A circular buffer of configurable size (in frames), subdivided into **periods**. Mixxx writes to this buffer in chunks, while the DMA engine drains it in `period_size` frames - a value negotiated between Mixxx and the audio-card driver at initialization. Usually, `period_size = Ring Buffer Size / 2`[^1]. Each time a period is emptied, ALSA triggers a software interrupt which is handled by a userspace callback in a high-priority Mixxx thread. This callback prepares frames and refills the ring-buffer.

**DAC FIFO** (hardware-level):  
   A small first-in-first-out queue that is read at the **DAC sample rate**, typically 44.1 kHz, 48 kHz, or 96kHz. This hardware buffer feeds the analog reconstruction circuitry with upstream frames. For instance, a 96kHz DAC consumes one frame every 1/96k s (≈10.4 us), totaling 96000 frames per second.

*Mixxx, ALSA, DAC buffers*
Mixxx Buffering Hierarchy

 While technically, only `period_size` frames are written to the DAC between callbacks via DMA, Mixxx prepares `Ring Buffer Size` frames in that duration. We can therefore simplify our model by noting that on average, `Ring Buffer Size` frames are written to the DAC every callback.
 
 ##### Buffer Underruns
From the `period_size` we compute `period_time = period_size / DAC Sample Rate`. This defines a hard real-time deadline for the userspace audio callback: it must prepare at least `period_size` frames within `period_time` to avoid starving the DAC’s hardware FIFO. This relation also confirms that large ring-buffers and lower DAC sample-rates reduce CPU pressure.

Whether this constraint is met depends on several factors - such as the complexity of audio processing in the real-time thread, OS scheduling latency, memory pressure, etc. Since general-purpose kernels do not provide any real-time guarantees, short `period_time` values can occasionally cause the callback to miss its deadline. The result is a **buffer underrun**, heard as a pop or glitch in playback—unacceptable in live DJ performance.

Unlike typical audio players, Mixxx performs real-time manipulation of audio—mixing, tempo changes, effects, scratching, and more - making the callback workload heavier. This demands low-latency implementations of all audio processing workflows to avoid underruns without compromising quality.

---

My intention was to give a generic description of the different samplerates (DAC, input), and how frames from a soundsource are consumed by the DAC. I am not sure what points are inaccurate. 
- For example, I think the calculations are correct, because ALSA behaviour (notify every `period_time`s) will be common regardless of the userspace buffering implementation. i.e. the mixxx DAC callback will be called every `period_time` rather than `Audio Buffer` time.

**Comment by daschuer** [comment]
I did a closer look into the buffering in Mixxx in this article: https://mixxx.org/news/2021-05-09-jack-zero-latency/ maybe it is helpful when describing the buffering here. 

Basically it works like this: 
* The Portaudio callback fires every buffer size (default 23 ms) and prepares pointers to two buffers per device https://github.com/mixxxdj/mixxx/blob/dabe2136b0e17f8cd158725ff27080e09f93a613/src/soundio/sounddeviceportaudio.cpp#L44
  * the recorded samples for input devices  
  * an empty buffer (containing old samples) which Mixxx has to filled. 
* With each callback Mixxx does this: 
  * Mixxx fetches the required samples form all playing audio files read ahead buffers 
  * Does stretching/resampling to device samplerate 
  * Applies effects, gain and mixing 
  * Mixxx need to finish before 23 ms is over (better at 10 ms or such for a bit of headroom for other task like visusscanning :-/ ) 
* Than Portaudio and the adio API like ALSA swap out the buffers. Ideally there are only two buffers involved (see link below) and feeds it to the DAC (via USB or a longer stack which also has buffering .. ) 
  * If Mixxx does not return it can't swap the buffers and the old one is feed again.

**Comment by daschuer** [comment]
New control commands can only be processed on whole buffers. So if you put the gain to zero for 10 ms and back it is ignored. 
If you want that it is processed you need to select a buffer size below 10 ms. This raises the CPU demand because every callback has a certain amount of management to do which needs to be done more often.

**Comment by daschuer** [comment]
> These parameters together influence the quality of output sound by defining the size in frames of the ALSA ring-buffer for the selected sample-rate:

Not exactly both do influence different properties: 

* Buffer size: Trade of CPU vs. responsiveness. (This is single core performance btw. which has not be enhanced so much in the past) 
* Sample rate: 
  * More is not equal better! Best is no resampling everything > 44 kHz can be treated as lossless.
  * 96 kHz wastes 50 % CPU power for processing unbearable frequencies.
  * Every resampling introduces noise.

**Comment by daschuer** [comment]
> Mixxx Buffering Hierarchy 

You probably have adopted you know ledge from other applications where your assumption applies. Mixxx is a low latency application which runs directly on a double buffer used in the DAC if the Harware allows. At least Portaudio has an abstraction layer that allows it, if the API allows. This assumption can be proved by a looback test from the linked article. 

You can think of it as a ring with two chunks. Mixxx filles the second half or the ring while the first halve has ben copied to the DAC.

**Comment by daschuer** [comment]
Interesting is the read-ahead-buffere (RAMAN). The resampler fatches the sample it needs from there. It is asynchronously filled with the most likely next required frames. Random jumps causes a zeroed buffere until samples heve been read. You can see it with audacity if you record such action.

**Comment by daschuer** [comment]
Does that make sense? Please ask for more detail if required. I must admit that I am probably operational blind to find a good way to explain it to our readers. 

I like also your hand drawings, a bit of personality in such a tech driven blog.

**Comment by Eve00000** [comment]
Would it be possible to add 2 audio fragments trying to represent the same action: scratching before (without your work)  & after (with your work on the the resampling issue) so users can hear the difference? Users should be able to hear the difference,  not?

**Comment by daschuer** [comment]
To be honest, I can't hear it. But our linear rescaler definitely adds measurable noise floor. There was a user that has shown nice spectral graphs about the issue. Maybe I find it. 
Maybe an expert is listening and can give some hints?

**Comment by daschuer** [comment]
@mxmilkiib It looks like you have some experience. (Taken from https://github.com/mixxxdj/mixxx/issues/7624#issuecomment-2920690077) Do you have interest to help here? 

With https://github.com/mixxxdj/mixxx/pull/15081
* Test 1: Play and record a 44.1 kHz track with 48 kHz hardware settings and Default recording sample rate and WAV encoding. The internal linear resample is used.   
* Test 2: Play and record a 44.1 kHz track with 44.1 kHz hardware settings and 48 kHz recording sample rate and WAV encoding. The libsamplerate with "Best quality" is used. 

Than listen to the differences and create something visually nice for this blog post.

**Comment by daschuer** [comment]
This is a nice one: https://therationalaudiophile.wordpress.com/2018/02/22/how-to-re-sample-a-signal/

Here is "something" tested https://src.infinitewave.ca/

**Comment by ar-in0** [comment]
> @mxmilkiib It looks like you have some experience. (Taken from [mixxxdj/mixxx#7624 (comment)](https://github.com/mixxxdj/mixxx/issues/7624#issuecomment-2920690077)) Do you have interest to help here?
> 
> With [mixxxdj/mixxx#15081](https://github.com/mixxxdj/mixxx/pull/15081)
> 
> * Test 1: Play and record a 44.1 kHz track with 48 kHz hardware settings and Default recording sample rate and WAV encoding. The internal linear resample is used.
> * Test 2: Play and record a 44.1 kHz track with 44.1 kHz hardware settings and 48 kHz recording sample rate and WAV encoding. The libsamplerate with "Best quality" is used.
> 
> Than listen to the differences and create something visually nice for this blog post.

I can do this, will use the gui scratch.

File: content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md [hunk 1]
```diff
@@ -1,17 +1,232 @@
...
+Mixxx uses the `SoundTouch` and `RubberBand` time-stretching libraries for resampling during a keyLock operation. However, these libraries are unsuitable for scratching due to the fast changing tempo and pitch. Currently a faster, handcrafted linear interpolation algorithm is used - but there have been reports of suboptimal audio quality.  
+
+Digital Signal Processing (DSP) theory tells us that linear interpolation is not ideal, and that a sinc-based resampler will always return interpolated values identical to the original analog signal, under certain theoretical constraints. However, practical implementations of sinc resampling are computationally heavy and generally unsuitable for low-latency realtime software such as Mixxx. Therefore,
+
+* One objective of this GSoC project was to explore the feasibility of using sinc interpolation for scratching. To this end, the `libsamplerate` and `libzita` resample latencies were evaluated.
+* Another focus was to investigate and improve the performance of the current linear resampler. Here, we observed that the `libsamplerate` linear interpolator outperformed our own, *reducing per-buffer resample latency from 20µs to 10µs.*
```
daschuer, line 19: Please use space between numbers and units See: https://physics.nist.gov/cuu/Units/checklist.html

daschuer:
Wanted to help with this: 

Time and Frequency Domains: 

In the time domain, resampling is done by combining nearby samples using a smart shaped weighting function, sample by sample as the time goes. In the frequency domain, a block of samples (window) is transformed (e.g. using an FFT) into its frequency and phase components (bins). These are shifted for the resampling operation and transformed back to time.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
+#### Theoretical Foundations
+Although of the work during GSOC involved integrating external C and C++ libraries into Mixxx, understanding the interpolation algorithms requires an overview of some foundational ideas in DSP.
 
-In the human ear, each cochlear hair cell is tuned to a specific frequency band and converts local mechanical vibrations into discrete neural spikes, encoding amplitude over time. This results in a time-series of electrical events that the brain interprets as sound.
+#### Keywords
+**Time and Frequency Domain:**  
 
-##### Periodic Sampling and Encoding
-While an analog signal is represented by its amplitude as continuous function in continuous time, the digital representation of that signal is a time series of amplitude values generated by noting the value of the analog signal at fixed, discrete time intervals.
+**Nyquist Rate:**
```
daschuer:
```suggestion
**Nyquist Frequency:**  

A digital (sampled) signal contains all frequencies up to the Nyquist frequency = half the samplerate (fs/2).
Everything below this frequency can be reproduced exactly. Frequencies above cannot be represented and must be removed before sampling to not cause aliasing noise. 
```

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
 
-##### Periodic Sampling and Encoding
-While an analog signal is represented by its amplitude as continuous function in continuous time, the digital representation of that signal is a time series of amplitude values generated by noting the value of the analog signal at fixed, discrete time intervals.
+**Nyquist Rate:**  
 
-*The process of generating a digital audio representation from an analog signal is termed **sampling***, and the length of the time interval is termed the "sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Mathematically, sampling is represented as multiplying an analog signal with an impluse train in the time-domain.
+**Periodic Sampling:** The process of generating a digital audio representation from an analog signal by recording amplitude at uniform intervals. The length of the time interval is termed the sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Analog audio is usually sampled at rates of 44.1KHz, 48Khz or 96Khz, ensuring the Nyquist criterion is met for frequencies in the audible range for humans `[20Hz - 20kHz]`
```
daschuer: I would move this to the top of the list.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
-While an analog signal is represented by its amplitude as continuous function in continuous time, the digital representation of that signal is a time series of amplitude values generated by noting the value of the analog signal at fixed, discrete time intervals.
+**Nyquist Rate:**  
 
-*The process of generating a digital audio representation from an analog signal is termed **sampling***, and the length of the time interval is termed the "sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Mathematically, sampling is represented as multiplying an analog signal with an impluse train in the time-domain.
+**Periodic Sampling:** The process of generating a digital audio representation from an analog signal by recording amplitude at uniform intervals. The length of the time interval is termed the sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Analog audio is usually sampled at rates of 44.1KHz, 48Khz or 96Khz, ensuring the Nyquist criterion is met for frequencies in the audible range for humans `[20Hz - 20kHz]`  
+**Digital Filter:**
```
daschuer: Works on sampled signals, often representing their analog counterparts. In digital form, the differential equations describing analog filters are replaced by discrete-time equivalents, which are implemented as convolution sums or, in block form, as matrix multiplications.

daschuer:
Wow it has already become real nice. Thank you for the extra efforts.

ar-in0:
> Wow it has already become real nice. Thank you for the extra efforts.

Thank you, will complete it by tonight.

ar-in0:
> Wanted to help with this:
> 
> Time and Frequency Domains:
> 
> In the time domain, resampling is done by combining nearby samples using a smart shaped weighting function, sample by sample as the time goes. In the frequency domain, a block of samples (window) is transformed (e.g. using an FFT) into its frequency and phase components (bins). These are shifted for the resampling operation and transformed back to time.

I wanted to avoid mentioning FFT/DFT because that is another involved topic. 
But yes if I add it I can add a section on frequency and time-domain resampling (to describe rubberband timestretch). 
If I have timebefore 11th, I will do it.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
+Mixxx uses the `SoundTouch` and `RubberBand` time-stretching libraries for resampling during a keyLock operation. However, these libraries are unsuitable for scratching due to the fast changing tempo and pitch. Currently a faster, handcrafted linear interpolation algorithm is used - but there have been reports of suboptimal audio quality.  
+
+Digital Signal Processing (DSP) theory tells us that linear interpolation is not ideal, and that a sinc-based resampler will always return interpolated values identical to the original analog signal, under certain theoretical constraints. However, practical implementations of sinc resampling are computationally heavy and generally unsuitable for low-latency realtime software such as Mixxx. Therefore,
```
daschuer:
```suggestion
Digital Signal Processing (DSP) theory tells us that linear interpolation is not ideal, and that a sinc-based resampler will always return interpolated values identical to the original analog signal, under certain theoretical constraints. However, practical implementations of sinc resampling are computationally heavy and generally unsuitable for low-latency realtime software such as Mixxx. Therefore ...
```

The three dots looks a bit better rendered to html.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
+**Nyquist Rate:**  
 
-*The process of generating a digital audio representation from an analog signal is termed **sampling***, and the length of the time interval is termed the "sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Mathematically, sampling is represented as multiplying an analog signal with an impluse train in the time-domain.
+**Periodic Sampling:** The process of generating a digital audio representation from an analog signal by recording amplitude at uniform intervals. The length of the time interval is termed the sample-period (its reciprocal - the **sampling rate** - is more commonly used while describing digital audio). Analog audio is usually sampled at rates of 44.1KHz, 48Khz or 96Khz, ensuring the Nyquist criterion is met for frequencies in the audible range for humans `[20Hz - 20kHz]`  
+**Digital Filter:**  
 
-*Periodic Sampling*
-![periodic sampling]({static}/images/news/mixxx-sampling-analog-digital.jpeg)
+**Impulse Response:**
```
daschuer: To measure properties of a filter a very short rectangular impulse signal is used as input containing all frequencies. The response of the filter is the output from this impulse signal, revealing frequency characteristics and group delay.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -1,52 +1,219 @@
...
+    <div style="width: 30%;">
+      <img src="{static}/images/news/mixxx-sinc-continuous.jpeg" alt="Periodic Sampling 1" style="width: 100%; height: auto;">
+      <p style="font-size: small; margin-top: 5px;"><strong>Figure 9:</strong> Sinc reconstruction kernel. In theory, convolving this function with a sample sequence implements a brick-wall filter.</p>
+    </div>
+  </div>
+</div>
+ <!-- In practice, we use N-taps of the sinc function, which implement a near-brick wall filter as N increases -->
+
+ To conclude:
+ > Convolving a sample sequence with a continuous sinc function is equivalent to applying a brick wall filter over its spectrum. In the time domain, this process results in reconstruction of the analog signal.
```
daschuer: I cannot follow this. 

My view is this (not sure if it is entirely correct) 

libsamplerate is a time domain function  by representing the signal as a sum of shifted sinc functions. For my understanding it expresses the signal as a number of sinc parts. This is done because they are describing a range of samples, including the values between two samples. For example between to peak samples of 0.8 was unlikely a flat top (a linear resampler assumes) but a sample of 0.85 or such. This can be recreated by resampling this sum of sinc forms with a different samplerate. 

Mabe you can show this issue as a drawing?

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
```diff
@@ -123,18 +290,46 @@ While buffer underruns occur due to DAC starvation regardless of the buffer cont
 
 In these scenarios, resampling is a corrective procedure that transforms audio sampled at the input sampling rate to match the DAC’s expected output rate.
 
-Resampling may also be used to *induce* tempo ramping. For a record sampled at 96kHz with a DAC also at 96kHz, scaling tempo by a factor of 3 means we want to pass 3x frames to the DAC on each callback than we would during standard playback.
+##### To induce tempo ramping
```
daschuer: Here a bief explaination is usefull. 

Mixxx acts on a buffer chunks eg. 20 ms. The scratch control command is taken into account between these buffers. To have a steady tempo without clicks and pops, a ramp is applied. If a user changes tempo from 1x to 2x. One buffer is used to slowly change the tempo from 1x to 2x.

daschuer:
I am through. Already a nice lecture.

daschuer:
> mention FFT 

Yes ithis topic is a bit misleading in your the resampler context.

content/news/2025-08-04-gsoc-2025-report-armaan-chowfin.md
ar-in0: >libsamplerate is a time domain function by representing the signal as a sum of shifted sinc functions. For my understanding it expresses the signal as a number of sinc parts. This is done because they are describing a range of samples, including the values between two samples. For example between to peak samples of 0.8 was unlikely a flat top (a linear resampler assumes) but a sample of 0.85 or such. This can be recreated by resampling this sum of sinc forms with a different samplerate.

Yes, this sounds correct. I could show the overlapping sincs in a diagram, (i was considering it already).

But I am unsure how libsamplerate is related to the  section you've highlighted in your github comment.
> Convolving a sample sequence with a continuous sinc function is equivalent to applying a brick wall filter over its spectrum. In the time domain, this process results in reconstruction of the analog signal.

daschuer: This is not related. I just want to verify if my view is correct. Your plan sounds good.

daschuer:
Wow, thank you. This will be hopefuly become a future reference for new contributors. LGTM.

PR #407, opened 2025-08-28
Fix/rework csp

Swiftb0y:
Images on https://mixxx.org/news/2025-08-06-qml-project/ got blocked because of accidental removal of the CSP to whitelist images loaded from our github org. The first commit re-adds that, ~~the second commit reformats the policy so changes like this are easier to spot in the future.~~ Unfortunately it doesn't seem to be possible to restructure the CSP directive in any way that's better documented. All approaches I've researched are non-standard.


JoergAtGithub (approved):
LGTM! Thank you for the fast fix!

PR #413, opened 2025-11-17
fix title of Zulip article

ywwg:


ywwg (commented):
my change also stripped trailing whitespace, which is also fine

daschuer (approved):
Thank you.

PR #416, opened 2025-12-09
Update press mirror

daschuer:
This adds two more links.


daschuer:
pre-commit fails because of German quotes.

ronso0:
Thanks!
pre-commit runs only on changed lines here as well, so this is a one-time issue, right?

ronso0 (approved):

daschuer:
> this is a one-time issue, right?

Yes.

PR #417, opened 2026-01-20
Credits drain fix 

daschuer:
This removes the big background images causing 67 % Bandwidth and enables client side caching. This shall save ~15 € per month. 

This is just a quick fix. On a long run we need a different solution.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -46,6 +46,23 @@
     X-XSS-Protection = "1; mode=block"
 
 [[headers]]
+  for = "/theme/fonts/*"
+  [headers.values]
+    Cache-Control = "public, max-age=31536000, immutable"
+    Access-Control-Allow-Origin = "*"
+    Vary = "Origin"
+
+[[headers]]
+  for = "/theme/css/*"
+  [headers.values]
+    Cache-Control = "public, max-age=31536000, immutable"
+
+[[headers]]
+for = "/theme/images/*"
+  [headers.values]
+    Cache-Control = "public, max-age=31536000, immutable"
+
+[[headers]]
  for = "/news/*"
  [headers.values]
    # Additionally allow YouTube/Discourse frames and scripts and MathJax
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -13,7 +13,7 @@ TIMEZONE = "Europe/Berlin"
 DEFAULT_LANG = "en"
 
 PATH = "content"
-FEED_ATOM = "feed.xml"
+EED_ATOM = None  # "feed.xml"
 
 ARTICLE_PATHS = [
     "news",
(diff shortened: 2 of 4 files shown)


ronso0:
No banner makes the website look a bit boring.
Re-encoding the splashscreen to jpeg 85% etc. would already reduce size to 10% (83 kB vs. 805 kB)
splash-1080x717

Eve00000:
The banner (and other static info like images) could also be placed elsewhere (not in netlify, eg downloadserver, or in a 2nd account)

daschuer:
I know all that. But can we just merge this without the banner and look if it really saves 1 € per day? 
If it turns out we are not able to reduce the credits spends near to the free threshold, all efforts for a better bandwidth safe solutions are mood. We will move to a unlimited Bandwidth host and fix the root cause, this is a better spend for our time IMHO.

JoergAtGithub (changes requested):
Tweaking the cache parameters is a good thing, but disabling the background image is not Ok. We can easily afford that little money until we switch to a webhoster with unlimated bandwidth.

daschuer:
We are talking about ~15 € a month for displaying backgrounds for eye candy only. Not a good spend IMHO.

acolombier:
I agree with @JoergAtGithub, tho I also agree we should find a solution not to give away this money. 
I find it impressive that there would be so much credits spent on this tho, has Netlify really become so expensive? Or is that just the result of AI crawler DDoS-ing the internet? 

I am wondering if adding a PoW like [Anubis](https://github.com/TecharoHQ/anubis) to prevent bot trafic, like Gnome did could help us reduce the traffic and bandwitdh

ronso0:
What about my proposal with better compression? Reduces bg image traffic to 10%
Owen also filed #418 bu twe shouldn't wait for someone to pick it up, just do it.

daschuer:
We had a discussion about this PR and decided to put it on hold until we have moved to a different hoster, because all these measures will unlikely push us under the free threshold.  We want to avoid any minimal regression in our service, and fix the root cause instead. @JoergAtGithub has agreed to become the "Kümmerer" for this, but he needs help. All the improvement Ideas can the be integrated to the new platform.

ronso0:
Sure, What I'm trying to say: if a few big images indeed cause unproportional traffic (compared site visits), then increasing compression (just that + the cache optimization) are a risk-free no-brainer IIUC?

daschuer:
That probably correct. But the issue is that I have no experience in such things and I like to spend my time elsewhere, given that we will move shortly to a new hoster where caching will be configured differently.

JoergAtGithub:
I think proper caching settings and compression (as long as it's invisible for the page viewer - which I assume for this blurry background image) are good anyway, as they reduce traffic and therefore electric energy consumption.
I would consider a slight compression for the background image risk free, as we can verify offline if it looks good on big screens or not. If this is given I can't imagine what else could break.
Caching settings are a different thing, where I can't rate the risk of breaking something.

daschuer:
The caching settings here are netlifly specific. Obsolete hopefully soon. If one want to take actions regarding the background, it is still usable later. We may even exchange them just for the sake of a fresh look after 5 years. 

See also 5627e1f5

daschuer:
This is a AI generated example picture, seasonal and more people focussed. What you can do with Mixxx, rather focussing the gear. How we can get such a picture? 
ChatGPT Image 24  Jan  2026, 13_06_15

ronso0:
Can we put this on hold now that the credits issue has been resolved (kinda)?
FWIW I think #418 is the better way forward

Regarding a new banner: I'd love to have a somewhat general image/setting/mood, that seasonal ice skating might make sense to us in "the north", but there's also the other hemisphere, and the tropics of course ; )

PR #432, opened 2026-06-06
25 Years of Mixxx DJ-Contest

Eve00000:
[preview](https://deploy-preview-432--mixxx-website.netlify.app/news/2026-08-06-25_years_of_mixxx_dj_contest/)


daschuer:
Already looks good. I am just am bit concerned if we find such a big number of participations. The entry bar is quite high, because of the Video/Youtube handling.
Do we want to allow also a file sharing uploads? Do we want to explicit describe that using a screen recorder is sufficient? 
Else looks already quite good.

acolombier:
This is a great idea! In terms of prize, would we be able to potentially put DJ gears to win? I am wondering if some manufacturers would consider making some gear available to us for the context, or offer discounted price.
This is a common practice that content creators leverage due to the visibility it offers to brand, I guess this case is no different?
Regarding the video, I agree with @daschuer that this is barrier for entry and also out of scope. I think we should require audio only submission. After all, Mixxx is an DJ software, and we have already expressed our position toward VJ not being part of the project. This feels like this would bias the challenge to allow video as part of the submission

Eve00000:
> ... if we find such a big number of participations. 

We do need to have some nice prizes ... I intentionally did not fill in values.
I would like to contact Phil for pub & contacts for prizes.

> ... Video/Youtube handling. Do we want to allow also a file sharing uploads? Do we want to explicit describe that using a screen recorder is sufficient?

We live in a video/tt/yt era .... if desired I can add an OBS manual ...

Eve00000:
> ... terms of prize, would we be able to potentially put DJ gears to win?

Agree, I would also prefer this or maybe we can refund a (part of a) purchase for the dedicated won prize-amount if the laureats buy a controller / soundcard .. and send us the receipt...

> ... making some gear available to us for the context, or offer discounted price. 

There could be the problem of delivery but I'd like to ask Phil how he's doing that with winners.

> ... require audio only submission. 

The video-idea:
- to inspire other users (benifit for the community)
- to be sure that the mix was made with Mixxx, it would be strange if someone won with a mix made with other software.
- to check the mix is authentic, made for the occasion and made by a human (yes yes)
- videos will attract much more voters / pub) than audio only 

We could allow screen recordings created with a phone, as long as the DJ-actions are clearly visible.

I thought / tested the processing of videos.
with the suggested workflow I can create the resulting video §intro/outro, + thumb & info captions) in 5-10 minutes, when I get  all kinds of codes generating the result can take up to 20-30 minutes per video... 
I suggested youtube uploads to avoid TT or other video services, YT will already convert all the videos (and audio)  to the same coded, so there won't be quality differences. If someone shares a video in a cloud services, there could be differences in audio.

daschuer:
I did not meant to allow audio-only submission. 
I just mean that many people do not instantly know how to upload a video to YouTube, some may even not want to use non FOSS tools. 

How about require just a screen recording, at a fixed resolution and codec, to make you work easy. 

In the finals we may extend it and require a mobile video of the performance along with the audio recorded in Mixxx, or anything else we need for promotion.

ronso0:
I very much like the idea of a mix contest -- but I'm also a bit worried about the video part.
Ideally it should not require any editing by us, assuming there'll be hundreds of submissions.

@Eve00000 how long do you think does it take Average Joanne/Joe to set up OBS (on win/mac/linux) and record a usable video (at native reolsution, ideally without post-processing)?
What's OBS' impact on CPU/Mixxx latency? Would it exclude users on low-spec machines?

Eve00000:
> I very much like the idea of a mix contest -- but I'm also a bit worried about the video part. Ideally it should not require any editing by us, assuming there'll be hundreds of submissions.

But if we want to make some pub with it: add our intro & outro, have some info about the DJ in a titlepage & subtitles (I have a kind of template), with the source from YT (so I don't need to do high/long CPU demanding renderings) I can convert a video from YT-download to ready in 5-10 minutes.

> .. hundreds ... 

that is why I suggested to limit it to 100.

> ...  Average Joanne/Joe to set up OBS (on win/mac/linux) and record a usable video ...

There are some nice tutorials like https://www.youtube.com/watch?v=xoe9ZOzlfnQ or  & https://www.youtube.com/watch?v=nQp0oA1CyzM
Tutorial for uploading to youtube:  https://www.youtube.com/watch?v=VtF2AgFSLAw

OBS is very very lightweighted (and we can accept people recording their screen with their phone if really needed)

Eve00000:
Extra:
Since the proposed schedule gives the DJs until September 30 to make their first video, there is sufficient time to experiment and, for example, experiment with  OBS and get a proper result with it.

Yes the Idea is to launch the Contest this month, so we can have 6 months of Pub :-)

acolombier:
> We live in a video/tt/yt era .... 

Do you mean to only include a screen cast of the Mixxx? I guess in this case, it could work. But most of all, we live in an era of AI, and I think it would be a shame if we endup with submission with VJ/visuals, which would bias the vote compare to a DJ-only submission.

> if desired I can add an OBS manual ...

Good luck with getting a manual that reliably works for all Linux user on Wayland :D

ronso0:
> we endup with submission with VJ/visuals, which would bias the vote compare to a DJ-only submission.

Good point! We want some kind of proof that the mix was created with Mixxx (eg. screen rec), but we also want to keep it fair for DJs that do proper mixes -- and are not visual artists.
So shall we restrict the video to screen recording then?

acolombier:
I think this is fair! We could allow pure screen recording, and screen recording + camera embedding, "Twitch-style" for those who want to also include they live performance with the controller.

Eve00000:
> So shall we restrict the video to screen recording then?

> We could allow pure screen recording, and screen recording + camera embedding, "Twitch-style" for those who want to also include they live performance with the controller.


My original idea was to have the Mixxx Gui fullscreen all the time / the whole video, as it was not clear for everybody
Can someone please make a suggestion to improve the formulation?
' The **Mixxx GUI must be visible in fullscreen** throughout the entire set so we can see (and check) your actions.'

Eve00000:
Another thing:
Imagine we contact vendors and we get some material to give to the top 3....
Vendors want something in return... banners, visibility, pub on our website / communication etc ...
And if we get material that we would personally not buy ... we'll need to metion it over and over

I would prefer to remain independent and have a neutral communication (not polutiog our website) and define the top 3 prizes as vouchers = we refund an amount when the winners send us the invoice of the DJ-Material they bought (Amazon, Bax, ... whatever DJ shop). I know it could cost us ssomething but independency has it's price.
What do you think?

daschuer:
It has another aspect: what if the winners wins something they have no use for? 

Getting in contact with a couple of vendors is IMHO a good idea independent form the DJ-Contest. The contest is however a good opportunity to do so. It may end up in a closer partnership and may establishes Mixxx more. Maybe a prototype supply that we have a Mixxx mapping ready on release of a new device or such. We have also discussed a "Mixxx Approved" label that we may want to establish. For all these things a good network to Hardware developer is useful. 

> I would prefer to remain independent and have a neutral communication

I can confirm that, we should not start to place adds or such or promote rubbish. 

Om the other hand it feels wrong to let the winners bye a device at the end of the value chain where we can also ask to get it directly from the vendor either as a free sample or at cost price. 

Is there a middle ground?

acolombier:
> Mixxx GUI must be visible in fullscreen

I think we should allow an exception of library because it may have private/personal/intimate info here (playlists name, file sources, or simply "mess"). I certainly wouldn't want to share mine.

> Vendors want something in return... banners, visibility, pub on our website / communication etc ...
And if we get material that we would personally not buy ... we'll need to metion it over and over

> we should not start to place adds or such or promote rubbish
> I can confirm that, we should not start to place adds or such or promote rubbish.

Not really how branded partnership works nowadays in the "influence" economy world, so I don't think there is such a worry to have. Today's currency is usually around a (or a few) post(s) so it is likely that we would have to "thanks" the vendor and share an affiliated link in the blog post introducing the challenge or the result, same in the Youtube video (if any). It's usually pretty ephemeral as well, so nothing prevents you from removing it after a "cooldown" period that we can request. 

> we refund an amount when the winners send us the invoice of the DJ-Material they bought

Honestly, I think this sounds too sketchy. There is also the fact that we have to spend our direct fund, which aren't much especially if we decide to do something else for the anniversary (e.g have a spot at audio.dev con or similar). 
I also feel wrong to giving donators money to US tech giants (Amazon meh), this would certainly make me think twice before making another Mixxx donation, but appreciate that last point is very personal.

Eve00000:
Thank you @acolombier for the insights and clearing up.
I prepared a letter/mail to contact the vendors.

acolombier:
Great! Happy to help as well with the reach.

acolombier (commented):
As I have previously expressed in internal channel, I believe that constraining the format of the submission beyond the time is directly against our main moto: DJ your way.

As I have been requested, I will stay out of the contest organisation and thus won't use my veto power here, however, note that I won't take part of the contest, though I am planning to publish a celebration set for the anniversary, which represent how I DJ my way.

daschuer:
The plans are looking good and well thought. Thank you for all that work. 

It is however quite a wall of text. Reading though all the rules, they probably take a bit away of the fun we may have.  
Can we do a and advertising Summary as the bog post and move the detailed rules to the wiki for example? Like our GSoC rules? 

What is clear to me, that it is hard to compare apples and pears. Therefore rules limiting "your way" do help here. 
This made the contest a fair challenge. Can be compared with "The Voice of ..." blind audition. The alternative would be a "... got talent" where any gimmick can be used to win over the audience. The can be felt as unfair when a 6 year old girl wins just because she is just 6 and so sweet.   

In therms of Mixxx as a equal technology platform, I can confirm that the original approach is most challenging and fair. But is probably also less fun as you can see in the cited TV shows. 

I am more if favour for the original "blind audition" mode, but maybe we can be more "your way" later on. 
Just my 2 cents ...

daschuer:
What is the state here. Is it ready to merge?

content/news/XXXX-XX-XX-25_years_of_mixxx_dj_contest.md
```diff
@@ -0,0 +1,283 @@
...
+comments: yes
+
+
+To celebrate the 25th anniversary of [Mixxx](https://mixxx.org) and its community, we want to organize an event where everyone can show how they use Mixxx, what tricks Mixxx makes possible, to demonstrate that Mixxx is designed to 'DJ your way' and how much fun we have with it.
+
+## Who?
+Everyone who loves music and DJing, whether you DJ for yourself, for friends and family or in public, whether you are a bedroom DJ, a tango DJ, a wedding DJ, a club DJ, or a Famous DJ playing on big events, whether you're a beginner or already experienced — everyone is invited to join.  
+Music is about emotions, DJing is about evoking emotions in others: with music you can make others happy, cheerful, enthusiastic or excited, but you can also make others melancholic, sad, or withdrawn.  
+
+If you prefer to share a 'Happy-Anniversary-Mixxx'-mix without participating in the DJ Contest or you want to share Anniversary-wishes, please do on the [forum](https://mixxx.discourse.group/), we will open a channel/topic there for the occasion.
```
daschuer, line 14: We should probably open the channel before posting this and link it here. Alternative would be to edit the post once we have the channel/topic.

Eve00000: I'll create a PR to link the topic when the news is online.

Eve00000:
> What is the state here. Is it ready to merge?

Yes, it was waiting ... first the T-Shirt news needed to be online.

daschuer:
Shall I merge now?

Eve00000:
> Shall I merge now?

yup yup, let's start .... exciting

daschuer:
Thank you for a ll the work you put into this.

Eve00000:
> Thank you for a ll the work you put into this.

correction: you -> we

PR #440, opened 2026-07-29
fix(OVH): remove accidental copy-paste leftover

acolombier:


JoergAtGithub:
Can you please explain, why the production condition is wrong here? Isn't the PR case not in the or condition?

acolombier:
The way ternary value works in GH CI is

```
<condition> && 'valueIfTrue' || 'valueIfFalse'
```

`production` is the `valueIsTrue` for the `environment_name` input value (which was accidentally copy pasted and missed by Copilot review). Technically, it still work because `&& 'production'` basically ends up being interpreted as a noop such that

```
<condition> && true && 'valueIfTrue' || 'valueIfFalse'
```

So this is really just noise and misleading, but doesn't have any impact per se

JoergAtGithub:
Ok, lets merge it!

PR #441, opened 2026-07-29
chore(OVH): delete environment when it gets tear down

acolombier:
Correctly clean up the active environment view: https://github.com/mixxxdj/website/deployments

### DIFF
--- a/.github/workflows/ovh.yaml
+++ b/.github/workflows/ovh.yaml
@@ -24,9 +24,12 @@ on:
         required: true
       OVH_PASSWORD:
         required: true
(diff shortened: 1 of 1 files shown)


acolombier:
Hmmm I was under the impression that `deployments: write` would be enough but on second look, it does need elevated permission, which only a PAT can grant.

I have set a dedicated PAT from @mixxxbot, I will grant it access once my access is sorted, as requested on Zulip.

acolombier:
Added @mixxxbot as admin on the project, should be good to go now!

PR #447, opened 2026-09-02
banner-idea

Eve00000:
[preview](https://deploy-preview-447--mixxx-website.netlify.app/)

This PR transforms the header of the website in a dynamic tectbanner

from
afbeelding

to 

afbeelding

and

afbeelding

and

afbeelding

and

afbeelding


daschuer:
Nice, looks really polished. It is in any case worth to keep. 

There is one empty topic (dot) though. 

I am just a bit concerned that people will not recognize it. If people scroll down a bit or click any of the links it is out of sight and they might missed it. So we need to do more.

Qt has a similar problem solved like this: 
https://www.qt.io/blog/qt-contributors-summit-2026-oslo-in-october
Probably to worth the effort to copy that 1:1, just to get the idea. 

What are low hanging fruits: 
Putting bar at topmost paragraph and just advertise the contest and T-shirts:  
image Or just place the 25th year medal their? 

Or adding two more buttons in the static header bar: Maybe with symbolic icons:  
image

image
For the contest we may use the 25th medal or such.

daschuer:
Oh I missed that the empty banner is this: 

image

netlify.toml
```diff
@@ -22,7 +22,8 @@
   [headers.values]
     # Only allow resources to be loaded from the current origin. Styles may
     # also be specified inline, and allow the Google Analytics scripts.
-    Content-Security-Policy = "default-src 'self'; style-src 'self'; script-src 'self'"
+    # Allow scripts from self and inline scripts for banner-rotation
+    Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'"
```
daschuer: I have no clue if there are implications with this. Who can check that?

daschuer: Do we even have inline scripts, or is this a leftover?

Eve00000:
> Nice, looks really polished. It is in any case worth to keep.

Thank you

 
> There is one empty topic (dot) though.

I was trying to get a png banner on the place of the main & subtitle (DJ Your Way & Free and open source DJ software for Windows, macOS, and Linux) but the banner always created more space between the logo & the text lines for text-only banners. 
So ... I dropped the png banners for now, I put a lot of time/enerfu to be as close as possible to the original while having dynamic messages. 
Logic: 5 secs homebanner -> 5 secs banner 1 -> 5 secs homebanner -> 5 secs banner 2 ->5 secs homebanner -> 5 secs banner 3 ->5 secs homebanner -> 5 secs banner 1 -> ....

> I am just a bit concerned that people will not recognize it. If people scroll down a bit or click any of the links it is out of sight and they might missed it. So we need to do more.

correct, we would solve that by having a fixed header ... ruining our design (which I like a lot)

netlify.toml
Eve00000: To be honest: I don't know, this is more the specialty of @acolombier & @Swiftb0y

Eve00000:
> Oh I missed that the empty banner is this:

It was working but taking to much place, the design with the Logo & the 2 baselines was ruined by having extra space between the logo & the baselines due to the png image banner ..... 

So I dropped the image banners.

Eve00000:
> What are low hanging fruits:
Putting bar at topmost paragraph and just advertise the contest and T-shirts:
Or just place the 25th year medal their?
Or adding two more buttons in the static header bar: Maybe with symbolic icons:

Let's have that in another PR

netlify.toml
daschuer: How about removing it and try if its still working?

theme/static/js/banner-rotation.js
```diff
@@ -0,0 +1,213 @@
...
+            {
+                type: "text",
+                main: "Version 2.5.6 available for download",
+                subtitle: "latest stable release",
+                url: "https://mixxx.org/download/",
+                target: "_self"
+            },
+            {
+                type: "text",
+                main: "25th Anniversary Artist T-Shirts available",
```
daschuer: Does this works? 
```suggestion
                main: "25<sup>th</sup> Anniversary Artist T-Shirts available",
```

theme/static/js/banner-rotation.js
```diff
@@ -0,0 +1,213 @@
...
+                subtitle: "order your exclusive T-Shirts now in our shop.",
+                url: "https://shop.mixxx.org",
+                target: "_blank"
+            },
+            {
+                type: "text",
+                main: "25th Anniversary Intergalactic DJ Contest",
```
daschuer:
```suggestion
                main: "25<sup>th</sup> Anniversary Intergalactic DJ Contest",
```

Eve00000:
afbeelding

nope, not html, needs to be unicode

netlify.toml
Eve00000: Did test & research .... yes we need them :-)

netlify.toml
```diff
@@ -22,7 +22,8 @@
   [headers.values]
     # Only allow resources to be loaded from the current origin. Styles may
     # also be specified inline, and allow the Google Analytics scripts.
-    Content-Security-Policy = "default-src 'self'; style-src 'self'; script-src 'self'"
+    # Allow scripts from self and inline scripts for banner-rotation
+	Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; frame-src 'self' https://app.netlify.com"
```
daschuer: Did you double check the 'unsafe-inline' flag? For my understanding we have no inline script code and it can be reverted. Otherwise name the required code here explicit in the comment.

Eve00000: Content-Security-Policy
- the added frame-src 'self' https://app.netlify.com" is for the white rectangle appearing on the bottom of the page in chrome .... appears to be a 'collaborate on this Deploy Preview' from netlify, searched for some time to get the error warning in the console solved ....

to make you happy: changing
Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; frame-src 'self' https://app.netlify.com"
to
Content-Security-Policy = "default-src 'self'; frame-src 'self' https://app.netlify.com"

results in all these errors:
afbeelding

afbeelding

afbeelding

daschuer: Thank you for testing.

daschuer (approved):
LGTM, thank you for this nice addition.

Eve00000:
Glad you like it... let's hope it attracts extra attention.

netlify.toml
acolombier: Just saw that - FYI, `script-src 'unsafe-inline'` is basically disabling XSS protection. In other terms, this exposes an visitor of `mixxx.org` to potential attack, in case mixxx.org was accidentally serving rogue contents. 

A quick prompt to Google or an LLM strongly recommand you not to do this so I would recommend reverting this change ASAP.

acolombier: CC @mixxxdj/developers

daschuer: For my understanding we don't have a risk here, because we do not display user data. So a revert of this PR is not rectified. However a solution without `unsafe-inline` is very welcome. Do you know one?

JoergAtGithub: I agree to @acolombier. Lets revert this until a safe solution is there!

JoergAtGithub: @mixxxdj/developers It is no longer possible to comment any blog post on our website, because mixxx.discourse.group rejects the connection:
grafik

ronso0: I see "Firefox can't open this page"

@Eve00000 can you provide a fix or a workaround?

Eve00000: I created https://github.com/mixxxdj/website/pull/453 in which I reverted the csp to the state before banner-idea was merged, but I can't see if the comments are working as they are only added in the real mixxx.org (discourse security), not in a deploy-preview @ netlify.app

PR #450, opened 2026-09-17
Update Homepage, add DJ Contest & T-Shirts Banner

Eve00000:
This pr contains an original idea from Daniel: put the prizetable on the homepage,
I added some DJ Contest info & a T-Shirts Banner

[preview](https://deploy-preview-450--mixxx-website.netlify.app/)


theme/templates/pages/index.html
```diff
@@ -59,6 +59,23 @@
 
 
 {% block body %}
+<div id="25thAnniversaryDJContest" class="docs-section container">
+  <h2><a class="headline-link" href="{{ SITEURL }}/news/2026-08-23-25_years_of_mixxx_dj_contest/">{% trans %}25ᵗʰ Anniversary Intergalactic DJ Contest{% endtrans %}</a></h2>
+  <p>{% trans %}Deadline to submit your '25 Years of Mixxx Intergalactic DJ Contest'-video is September 30ᵗʰ 18:00 CET.<br>
+  DJs of all styles, genres and skill levels are welcome to participate.<br>
+  If you are stuck on recording your set on video and need some help, find us on the <a href="https://mixxx.discourse.group/t/25-years-of-mixxx-intergalactic-dj-contest/34141/26" target="_blank" rel="noopener noreferrer">Mixxx-forum</a>{% endtrans %}</p>
+  <a href="{{ SITEURL }}/images/news/djcontest_prizetable.png">
```
ronso0: I expected this to bring me to the linked page, not the PNG
Same for the T-Shirts banner

ronso0 (commented):
Thanks for the update!
Though I'm not sure I like this being at the top of the landing page, so I'll just leave a comment on the UX

theme/templates/pages/index.html
Eve00000: Ah ok,
at the moment the title is linking to the pages, same behaviour as the 'features ... ' and the images link to the big version.

I'll try to update.

Eve00000: > Though I'm not sure I like this being at the top of the landing page

I have the same sentiment

ronso0: > and the images link to the big version.

exactly, and my expectation from blogs' overviews/timelines is that clicking the title (text) or the image brings me to the article. If I want to see the image I explicitly open it in a new tab with the context menu.

Eve00000: adjusted

ronso0:
Thank you, it works now.
@daschuer is this what you had in mind?

daschuer:
> @daschuer is this what you had in mind?

Actually yes. But I now share you concerns. 

> Though I'm not sure I like this being at the top of the landing page, so I'll just leave a comment on the UX

We need probably a better idea. 

Brainstorming: 

Even without the advertisements our Landing page does not work well. 

Looking here: 
https://www.marshall.com
https://www.reloop.com/

They have the main product photo on the top with "buy now" call to action button. 
The merch and events follow below. 


They have a sticky side note that moves: 
https://www.qt.io/blog/qt-contributors-summit-2026-oslo-in-october


The low hanging fruit would be here just swap the order. 

Do we want a screenshot of Mixxx on th main page? 


The other issue I see is that the Price table looks a bit too detailed for my eyes. Maybe we can join both articles and make banner based on the t-shirts banner but with only the monochrome logos of our sponsors and a monochome t-shirt?

